PR #3, opened 2022-06-30
Max oo

rileyjmurray:


src/comps/qb.cc
```diff
@@ -0,0 +1,407 @@
...
+        {
+            gemm(Layout::ColMajor, Op::Trans, Op::NoTrans,
+                k, k, m,
+                1.0, Q_dat, m, Q_dat, m,
+                0.0, Q_gram_dat, k
+            );
+            for (int oi = 0; oi < next_sz; ++oi) {
+                Q_gram_dat[oi * k + oi] -= 1.0;
+            }
+            T orth_err = lange(Norm::Fro, k, k, Q_gram_dat, k);
```
rileyjmurray, line 174: It would be nice to only write to the first ``next_sz * next_sz`` entries of ``Q_gram_dat``. We can just lie about it's leading dimension.

src/comps/qb.cc
```diff
@@ -0,0 +1,407 @@
...
+    
+            if(QB::verbosity)
```
rileyjmurray, line 176: Use ``this->``

src/comps/qb.cc
```diff
@@ -0,0 +1,407 @@
...
+        
+        // This step is only necessary for the next iteration
+        // A = A - Q_i * B_i
+        gemm(Layout::ColMajor, Op::NoTrans, Op::NoTrans, m, n, block_sz, -1.0, Q_i_dat, m, B_i_dat, block_sz, 1.0, A_cpy_dat, m);
+
+        //char name_5[] = "A_updated";
+        //RandBLAS::util::print_colmaj(m, n, A_cpy_dat, name_5);
+    }
+    // Reached expected rank without achieving the tolerance
+    QB::termination = 3;
```
rileyjmurray, line 210: Use ``this->termination``

src/comps/qb.cc
```diff
@@ -0,0 +1,407 @@
...
+void QB<T>::QB2_test_mode(
+    int64_t m,
+    int64_t n,
+    std::vector<T>& A,
+    int64_t& k,
+    int64_t block_sz,
+    T tol,
+    std::vector<T>& Q,
+    std::vector<T>& B
+){
```
rileyjmurray, line 223: My understanding is that this function is almost identical to ``QB2``. You should have comments that indicate where the differences are.

src/comps/rf.cc
```diff
@@ -0,0 +1,176 @@
...
+    std::vector<T>& Q, // n by k
+    bool use_qr
+){
+    printf("HI, I AM REGULAR RF. \n");
+
+    using namespace blas;
+    using namespace lapack;
+
+    // Get the sketching operator Omega
+    std::vector<T> Omega(n * k, 0.0);
```
rileyjmurray, line 26: Should store ``Omega`` is a state variable and resize it if necessary.

src/comps/rf.cc
```diff
@@ -0,0 +1,176 @@
...
+    T* Q_dat = Q.data();
+
+    RF::RS_Obj.call(m, n, A, k, Omega);
```
rileyjmurray, line 29: As with my feedback for the QB code, make sure you use ``this->`` appropriately (for the most part it should be used instead of ``RF::`` in this file).

src/comps/rf.cc
```diff
@@ -0,0 +1,176 @@
...
+    gemm<T>(Layout::ColMajor, Op::NoTrans, Op::NoTrans, m, k, n, 1.0, A.data(), m, Omega.data(), n, 0.0, Q_dat, m);
+
+
+    //char name_5[] = "Gemm result";
+    //RandBLAS::util::print_colmaj(m, k, Q_dat, name_5);
+
+    if (RF::cond_check)
+    {
+        // Copy to avoid any changes
+        std::vector<T> Q_cpy (m * k, 0.0);
```
rileyjmurray, line 41: Like ``Omega``, this should also be a state variable.

src/comps/rf.cc
```diff
@@ -0,0 +1,176 @@
...
+        std::vector<T> s(k, 0.0);
+        T* Q_cpy_dat = Q_cpy.data();
+        T* s_dat = s.data();
+
+        lacpy(MatrixType::General, m, k, Q_dat, m, Q_cpy_dat, m);
+        gesdd(Job::NoVec, m, k, Q_cpy_dat, m, s_dat, NULL, m, NULL, k);
+        T cond_num = *s_dat / *(s_dat + k - 1);
```
rileyjmurray, line 48: I'm not a fan of this pointer arithmetic. It would be better to just do ``s_dat[0] / s_dat[k - 1]``.

src/comps/rf.cc
```diff
@@ -0,0 +1,176 @@
...
+
+template <typename T>
+void RF<T>::rf1_test_mode(
+    int64_t m,
+    int64_t n,
+    const std::vector<T>& A,
+    int64_t k,
+    std::vector<T>& Q,
+    T& cond_num
+){
```
rileyjmurray, line 98: As with my feedback for QB, you should explain how ``rf1_test_mode`` compares to ``rf1``.

src/comps/rs.cc
```diff
@@ -0,0 +1,141 @@
...
+	const T* A_dat = A.data();
+	T* Omega_dat = Omega.data();
+	T* Omega_1_dat = Omega_1.data();
+
+	if (p % 2 == 0) 
+	{
+		// Fill n by k Omega
+		RandBLAS::dense_op::gen_rmat_norm<T>(n, k, Omega_dat, seed);
+		return;
+	}
```
rileyjmurray, line 59: We shouldn't return unless ``p == 0``.

src/comps/util.cc
```diff
@@ -0,0 +1,435 @@
...
+        int64_t k
+) {     
+        using namespace blas;
+        T* A_dat = A.data();
+        uint64_t end = k;
+
+        for (int i = 1; i < n; ++i)
+        {
+                // Place jth column (of k entries) after the (j - 1)st column
+                copy(k, A_dat + (m * i), 1, A_dat + end, 1);
```
rileyjmurray, line 130: No pointer arithmetic.

src/comps/util.cc
```diff
@@ -0,0 +1,435 @@
...
+                        }
+                ); 
+                // The unit diagonal elements of L were not stored.
+                L_dat[i + j] = 1;
+        }
+}
+
+// Addressing Pivoting
+template <typename T> 
+void row_swap(
```
rileyjmurray, line 96: I'd call this ``swap_rows``, with a plural.

rileyjmurray (commented):
ping @TeachRaccooon. A lot of comments here for ya. Here are some recurring things:
1. Don't allocate workspace matrices from scratch every time an algorithm is called. Have the algorithm object maintain workspace as state variables (std::vector) and resize as necessary.
2. I'd like to avoid ``using namespace blas`` or ``using namespace lapack``. I think it makes code harder to read for non-experts.
3. Don't use pointer arithmetic. If you need a pointer to a location ``i`` places after ``ptr``, then use ``&ptr[i]``.

.gitmodules
```diff
@@ -1,3 +1,6 @@
 [submodule "RandBLAS"]
 	path = RandBLAS
 	url = https://github.com/BallisticLA/RandBLAS
+[submodule "Submod"]
+	url = https://github.com/BallisticLA/RandBLAS/tree/max_branch
+	branch = max_branch
```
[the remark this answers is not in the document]
TeachRaccooon: ¯\_(ツ)_/¯

src/comps/qb.cc
```diff
@@ -76,6 +81,7 @@ void QB<T>::QB2(
         block_sz = std::min(block_sz, k - curr_sz);
         int next_sz = curr_sz + block_sz;
 
+        ++QB::RF_Obj.RS_Obj.seed;
```
rileyjmurray, line 84: This probably won't be sufficient. It should be incremented by [number of generated numbers] after each call to the counter-based RNG.

langou (commented):
Cool. Go RFP!

benchmark/CMakeLists.txt
```diff
@@ -0,0 +1,12 @@
...
+set(tmp FALSE)
+if (GTest_FOUND)
+    set(tmp TRUE)
+
+    add_executable(RandLAPACK_benchmark experiments/comps/QB_cond_nums.cc experiments/comps/Orth_speed.cc)
+    target_link_libraries(RandLAPACK_benchmark OpenMP::OpenMP_CXX RandLAPACK GTest::GTest GTest::Main)
+    gtest_discover_tests(RandLAPACK_benchmark)
+
+endif()
+message(STATUS "Checking for benchmarks ... ${tmp}")
```
rileyjmurray, line 12: Requiring GTest for benchmarks is fine for now, but we don't want that long-term.

benchmark/experiments/comps/Orth_speed.cc
```diff
@@ -0,0 +1,253 @@
...
+                long t_lu   = 0;
+                long t_qr   = 0;
+                long t_geqr = 0;
+
+                long curr_t_chol = 0;
+                long curr_t_lu   = 0;
+                long curr_t_qr   = 0;
+                long curr_t_geqr = 0;
+
+                std::ofstream file("../../build/test_plots/test_speed/raw_data/test_" + std::to_string(rows) + "_" + std::to_string(cols) + ".dat");
```
rileyjmurray, line 124: If we use relative file paths then we'll want to make sure C++ doesn't fail silently. That is, we'll want to raise an error if a file doesn't exist. But we can probably come back to this later.

benchmark/experiments/comps/Orth_speed.cc
```diff
@@ -0,0 +1,253 @@
...
+                    res = test_speed_helper<T>(rows, cols, 1);
+                    curr_t_chol = std::get<0>(res);
+                    curr_t_lu   = std::get<1>(res);
+                    curr_t_qr   = std::get<2>(res);
+                    curr_t_geqr = std::get<3>(res);
+
+                    // Skip first iteration, as it tends to produce garbage results
+                    if (i != 0)
+                    {
+
```
rileyjmurray, line 213: Remove this blank line.

include/RandLAPACK/comps/qb.hh
```diff
@@ -0,0 +1,105 @@
...
+template <typename T>
+class QBalg
+{
+        public:
+                virtual void call(
+                        int64_t m,
+                        int64_t n,
+                        std::vector<T>& A,
+                        int64_t& k,
+                        int64_t block_sz,
```
rileyjmurray, line 23: Block size isn't part of the call function for ``QBDecomposer`` in Python. Did we decide to keep it here out of an idea that unblocked QB isn't worth bothering with?

include/RandLAPACK/comps/qb.hh
```diff
@@ -0,0 +1,105 @@
...
+template <typename T>
+class QB : public QBalg<T>
+{
+	public:
+                RandLAPACK::comps::rf::RangeFinder<T>& RF_Obj;
+                RandLAPACK::comps::orth::Stabilization<T>& Orth_Obj;
+                bool verbosity;
+                bool orth_check;
+
+                // Avoiding preallocations
```
rileyjmurray, line 39: I don't understand what this comment is getting at. Based on what I see, I'd call this "declare (but do not allocate) workspace".

include/RandLAPACK/comps/qb.hh
```diff
@@ -0,0 +1,105 @@
...
+                        std::vector<T>& B
+                )
+                {
+                        switch(this->decision_QB)
+                        {
+                                case 0:
+                                        QB2(m, n, A, k, block_sz, tol, Q, B);
+                                        break;
+                        }
+                }
```
rileyjmurray, line 102: I don't see a reason to define ``QB2`` as a separate function in this context. (If this class also had ``QB3`` that would be a different story.)

include/RandLAPACK/comps/qb.hh
```diff
@@ -0,0 +1,105 @@
...
+};
+#endif
+} // end namespace RandLAPACK::comps::rs
```
rileyjmurray, line 105: Wrong namespace, I think.

include/RandLAPACK/comps/rs.hh
```diff
@@ -0,0 +1,88 @@
...
+			switch(this->decision_RS)
+			{
+				case 0:
+					rs1(m, n, A, k, Omega);
+					break;
+			}
+		}
+};
+#endif
+} // end namespace RandLAPACK::comps::rs
```
rileyjmurray, line 88: Wrong namespace

include/RandLAPACK/comps/util.hh
```diff
@@ -0,0 +1,134 @@
...
+        std::vector<T>& S
+);
+
+template <typename T> 
+void disp_diag(
+        int64_t m,
+        int64_t n,
+        int64_t k, 
+        std::vector<T>& A 
+);
```
rileyjmurray, line 37: Is ``k`` the number of elements to display, or the offset of the diagonal (e.g., ``k=1`` means first superdiagonal, ``k=-2`` means second subdiagonal)?

include/RandLAPACK/comps/util.hh
```diff
@@ -0,0 +1,134 @@
...
+
+template <typename T> 
+void swap_rows(
+        int64_t m,
+        int64_t n,
+        std::vector<T>& A, // pointer to the beginning
```
rileyjmurray, line 43: This comment ``//pointer to the beginning`` isn't needed.

include/RandLAPACK/comps/util.hh
```diff
@@ -0,0 +1,134 @@
...
+        int64_t n,
+        std::vector<T>& A
+);
+
+template <typename T> 
+void get_L(
+        int64_t m,
+        int64_t n,
+        std::vector<T>& L
+);
```
rileyjmurray, line 20: Write a short comment saying what this does.

include/RandLAPACK/comps/util.hh
```diff
@@ -0,0 +1,134 @@
...
+        int64_t n,
+        std::vector<T>& A, // pointer to the beginning
+        const std::vector<int64_t>& p // Pivot vector
+);
+
+template <typename T> 
+T* resize(
+        int64_t target_sz,
+        std::vector<T>& A
+);
```
rileyjmurray, line 51: I'd rename to ``upsize``, and provide a short comment explaining what this function does.

include/RandLAPACK/comps/util.hh
```diff
@@ -0,0 +1,134 @@
...
+template <typename T> 
+void gen_exp_mat(
+        int64_t& m,
+        int64_t& n,
+        std::vector<T>& A,
+        int64_t k, // vector length
+        T t, // controls the decay. The higher the value, the faster the decay
+        bool diagon,
+        int32_t seed
+);
```
rileyjmurray, line 91: Add comments explaining what these functions do.

src/comps/orth.cc
```diff
@@ -0,0 +1,115 @@
...
+#include <RandLAPACK.hh>
+
+namespace RandLAPACK::comps::orth {
+
+// Perfoms a Cholesky QR factorization
+template <typename T> 
+int Orth<T>::CholQR(
+        int64_t m,
+        int64_t k,
+        std::vector<T>& Q // pointer to the beginning
```
rileyjmurray, line 17: I'm going to stop pointing this out, but ``// pointer to the beginning'' comments aren't needed in general and should be used sparingly.

src/comps/orth.cc
```diff
@@ -0,0 +1,115 @@
...
+){
+        using namespace blas;
+        using namespace lapack;
+        
+        T* Q_gram_dat = RandLAPACK::comps::util::resize(k * (k - 1) / 2, this->Q_gram);
```
rileyjmurray, line 22: Don't you need ``k * (k + 1)/2``?

src/comps/orth.cc
```diff
@@ -0,0 +1,115 @@
...
+        int64_t n,
+        std::vector<T>& A,
+        std::vector<T>& tvec
+){
+        using namespace lapack;
+
+        tvec.resize(5);
+
+        T* A_dat = A.data();
+        T* tvec_dat = tvec.data();
```
rileyjmurray, line 94: There's no advantage to defining ``tvec_dat`` here since it's only used in one function call. Just use ``tvec.data()`` in that function call instead.

src/comps/orth.cc
```diff
@@ -0,0 +1,115 @@
...
+        
+        geqr(m, n, A_dat, m, tvec_dat, -1);
+        int64_t tsize = (int64_t) tvec[0]; 
+        tvec.resize(tsize);
+        geqr(m, n, A_dat, m, tvec.data(), tsize);
+
+        return 0;
```
rileyjmurray, line 101: There should be a check for whether ``geqr`` succeeded or returned an error code.

src/comps/qb.cc
```diff
@@ -0,0 +1,194 @@
...
+#include <RandBLAS.hh>
+#include <lapack.hh>
+#include <RandLAPACK.hh>
+
+#include <math.h>
+
+#define ORTHONORM_CHECKS
+#define COND_CHECK
+#define VERBOSE
+#define USE_QR
```
rileyjmurray, line 20: I don't think you use these anymore.

src/comps/qb.cc
```diff
@@ -0,0 +1,194 @@
...
+        k = curr_sz;
+        this->termination = 1;
+        return;
+    }
+    
+    // If the space allocated for col in Q and row in B is insufficient for any iterations ...
+    if(std::max( Q.size() / m, B.size() / n) < k)
+    {
+        // ... allocate more!
+        this->curr_lim = std::min(4 * block_sz, k);
```
rileyjmurray, line 57: Instead of using 4 as a hard-coded magic number, how about you define it as an instance variable and set it to 4 by default. (In my opinion we wouldn't even need to expose that instance variable in the constructor.)

src/comps/qb.cc
```diff
@@ -0,0 +1,194 @@
...
+
+    if(this->orth_check)
+    {
+        Q_gram_dat = RandLAPACK::comps::util::resize(this->curr_lim * this->curr_lim, this->Q_gram);
+        Q_i_gram_dat = RandLAPACK::comps::util::resize(block_sz * block_sz, this->Q_i_gram);
+    }
+
+    T* QtQi_dat = RandLAPACK::comps::util::resize(this->curr_lim * block_sz, this->QtQi);
+    T* Q_i_dat = RandLAPACK::comps::util::resize(m * block_sz, this->Q_i);
+    T* B_i_dat = RandLAPACK::comps::util::resize(block_sz * n, this->B_i);
```
rileyjmurray, line 87: These calls to resize are really verbose. Is there a C++ equivalent of a "Pythonic" ``import randlapack.comps.util as rlautil``?

src/comps/qb.cc
```diff
@@ -0,0 +1,194 @@
...
+        next_sz = curr_sz + block_sz;
+
+        // Make sure we have enough space for everything
+        if(next_sz > this->curr_lim)
+        {
+            this->curr_lim = std::min(2 * this->curr_lim, k);
+            Q_dat = RandLAPACK::comps::util::resize(this->curr_lim * m, Q);
+            B_dat = RandLAPACK::comps::util::row_resize(curr_sz, n, B, this->curr_lim);
+            QtQi_dat = RandLAPACK::comps::util::resize(this->curr_lim * block_sz, QtQi);
+            Q_gram_dat = RandLAPACK::comps::util::resize(this->curr_lim * this->curr_lim, Q_gram);
```
rileyjmurray, line 105: No need to resize ``Q_gram`` unless ``this->orth_check`` is true.

src/comps/qb.cc
```diff
@@ -0,0 +1,194 @@
...
+        }
+
+        // Calling RangeFinder
+        this->RF_Obj.call(m, n, A_cpy, block_sz, this->Q_i);
+
+        if(this->orth_check)
+        {
+            if (RandLAPACK::comps::util::orthogonality_check(m, block_sz, block_sz, Q_i, Q_i_gram, this->verbosity))
```
rileyjmurray, line 113: Same comment as before: can we make this function call more less verbose by an appropriate "import" statement?

src/comps/qb.cc
```diff
@@ -0,0 +1,194 @@
...
+        // No need to reorthogonalize on the 1st pass
+        if(curr_sz != 0)
+        {
+            // Q_i = orth(Q_i - Q(Q'Q_i))
+            gemm(Layout::ColMajor, Op::Trans, Op::NoTrans, curr_sz, block_sz, m, 1.0, Q_dat, m, Q_i_dat, m, 0.0, QtQi_dat, this->curr_lim);
+            gemm(Layout::ColMajor, Op::NoTrans, Op::NoTrans, m, block_sz, curr_sz, -1.0, Q_dat, m, QtQi_dat, this->curr_lim, 1.0, Q_i_dat, m);
+
+            // If CholQR failed, will use HQR
+            if(this->Orth_Obj.call(m, block_sz, this->Q_i))
+                this->Orth_Obj.call(m, block_sz, this->Q_i);
```
rileyjmurray, line 132: This is making an assumption about the semantics of ``Orth_Obj.call(...)``. Specifically, it assumes that a stateful variable is set so that a second call to the same function can produce different results. My preference would be for ``Orth_Obj.call`` to be responsible for falling back on a backup algorithm if its primary algorithm failed.

src/comps/qb.cc
```diff
@@ -0,0 +1,194 @@
...
+
+        //B_i = Q_i' * A
+        gemm<T>(Layout::ColMajor, Op::Trans, Op::NoTrans, block_sz, n, m, 1.0, Q_i_dat, m, A_cpy_dat, m, 0.0, B_i_dat, block_sz);
+
+        // Updating B norm estimation
+        T norm_B_i = lange(Norm::Fro, block_sz, n, B_i_dat, block_sz);
+        norm_B = hypot(norm_B, norm_B_i);
+        // Updating approximation error
+        prev_err = approx_err;
+        approx_err = sqrt(abs(norm_A - norm_B) * (norm_A + norm_B)) / norm_A;
```
rileyjmurray, line 143: Re-arrange: ``approx_err = sqrt(abs(norm_A - norm_B)) * (sqrt(norm_A + norm_B) / norm_A);``

(This alternative way of drawing parenthesis ensures that the the potentially-small number ``abs(norm_A - norm_B)`` isn't multiplied by a very large number ``(norm_A + norm_B)``.)

src/comps/qb.cc
```diff
@@ -0,0 +1,194 @@
...
+            // Early termination - error growth
+            // Only need to move B's data, no resizing
+            RandLAPACK::comps::util::row_resize(this->curr_lim, n, B, curr_sz);
+            k = curr_sz;
+            this->termination = 2;
+            return;
+        } 
+
+        // Update the matrices Q and B
+        lacpy(MatrixType::General, m, block_sz, Q_i_dat, m, Q_dat + (m * curr_sz), m);
```
rileyjmurray, line 157: Don't use pointer arithmetic.

src/comps/qb.cc
```diff
@@ -0,0 +1,194 @@
...
+            // Only need to move B's data, no resizing
+            RandLAPACK::comps::util::row_resize(this->curr_lim, n, B, curr_sz);
+            k = curr_sz;
+            this->termination = 2;
+            return;
+        } 
+
+        // Update the matrices Q and B
+        lacpy(MatrixType::General, m, block_sz, Q_i_dat, m, Q_dat + (m * curr_sz), m);	
+        lacpy(MatrixType::General, block_sz, n, B_i_dat, block_sz, B_dat + curr_sz, this->curr_lim);
```
rileyjmurray, line 158: No pointer arithmetic.

src/comps/qb.cc
```diff
@@ -0,0 +1,194 @@
...
+        
+        if(this->orth_check)
+        {
+            if (RandLAPACK::comps::util::orthogonality_check(m, this->curr_lim, next_sz, Q, Q_gram, this->verbosity))
+            {
+                // Lost orthonormality of Q
+                RandLAPACK::comps::util::row_resize(this->curr_lim, n, B, curr_sz);
+                k = curr_sz;
+
```
rileyjmurray, line 167: Remove blank line.

src/comps/rf.cc
```diff
@@ -0,0 +1,40 @@
+#include <lapack.hh>
+#include <RandBLAS.hh>
+#include <RandLAPACK.hh>
+
+#define USE_QR
+#define COND_CHECK
+#define VERBOSE
```
rileyjmurray, line 7: Remove these.

src/comps/rf.cc
```diff
@@ -0,0 +1,40 @@
...
+    T* Omega_dat = RandLAPACK::comps::util::resize(n * k, this->Omega);
+    T* Q_dat = Q.data();
+
+    this->RS_Obj.call(m, n, A, k, this->Omega);
+
+    // Q = orth(A * Omega)
+    gemm<T>(Layout::ColMajor, Op::NoTrans, Op::NoTrans, m, k, n, 1.0, A.data(), m, Omega_dat, n, 0.0, Q_dat, m);
+
+    if(this->cond_check)
+        RandLAPACK::comps::util::cond_num_check(m, k, Q, this->Q_cpy, this->s, this->cond_nums, this->verbosity);
```
rileyjmurray, line 31: What is this doing? Is it writing to ``this->cond_nums``?

src/comps/rf.cc
```diff
@@ -0,0 +1,40 @@
...
+    
+    // If CholQR failed, will use HQR
+    if(this->Orth_Obj.call(m, k, Q))
+        this->Orth_Obj.call(m, k, Q);
```
rileyjmurray, line 35: Same comment as before: this assumes semantics of ``Orth_Obj`` where a second call uses a different algorithm if the first call returns a nonzero number. I think that short of thing is better handled internally by ``Orth_Obj``.

src/comps/rs.cc
```diff
@@ -0,0 +1,81 @@
...
+	
+	while (p - p_done > 0) 
+	{
+		// Omega = A * Omega
+		gemm<T>(Layout::ColMajor, Op::NoTrans, Op::NoTrans, m, k, n, 1.0, A_dat, m, Omega_dat, n, 0.0, Omega_1_dat, m);
+		++ p_done;
+		if (p_done % q == 0) 
+		{
+			if(this->Stab_Obj.call(m, k, Omega_1))
+				this->Stab_Obj.call(m, k, Omega_1);
```
rileyjmurray, line 57: ``Stab_Obj`` should internally make the decision to fall back on another implementation in the event of an error.

src/comps/rs.cc
```diff
@@ -0,0 +1,81 @@
...
+		if(this->cond_check)
+			RandLAPACK::comps::util::cond_num_check(m, k, Omega_1, this->Omega_1_cpy, this->s, this->cond_nums, this->verbosity);
+
+		// Omega = A' * Omega
+		gemm<T>(Layout::ColMajor, Op::Trans, Op::NoTrans, n, k, m, 1.0, A_dat, m, Omega_1_dat, m, 0.0, Omega_dat, n);
+		++ p_done;
+		if (p_done % q == 0) 
+		{
+			if(this->Stab_Obj.call(n, k, Omega))
+				this->Stab_Obj.call(n, k, Omega);
```
rileyjmurray, line 69: ``Stab_Obj`` should  make the decision for whether or not to do this.

test/src/comps/test_util.cc
```diff
@@ -0,0 +1,26 @@
...
+    virtual void SetUp() {};
+
+    virtual void TearDown() {};
+
+    template <typename T>
+    static void 
+    test_1()
+    {
+    }
+};
```
rileyjmurray, line 26: Three options:

1. Add tests.
2. Write ``TODO: add tests``.
3. Delete this file.

src/comps/rs.cc
```diff
@@ -0,0 +1,81 @@
...
+		}
+		
+		if(this->cond_check)
+			RandLAPACK::comps::util::cond_num_check(n, k, Omega, this->Omega_cpy, this->s, this->cond_nums, this->verbosity);
```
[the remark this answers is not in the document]
TeachRaccooon: Probably no sense in that, as I shortened it by "using namespace."

include/RandLAPACK/comps/qb.hh
TeachRaccooon: Note in TODO taken.

benchmark/experiments/comps/Orth_speed.cc
TeachRaccooon: Note in TODO taken

PR #10, opened 2022-12-07
WIP -- For ptn integrated pipeline staging 2

burlen:
this depends on https://github.com/BallisticLA/RandBLAS/pull/17 TODO: update the submodule once 17 is merged


src/CMakeLists.txt
```diff
@@ -5,25 +5,25 @@ set(RandLAPACK_cxx_srcs
...
+target_compile_options(RandLAPACK PUBLIC -Wall -Wextra)
 
 target_include_directories(
     RandLAPACK PUBLIC
-    $<BUILD_INTERFACE:${CMAKE_SOURCE_DIR}/include>
-    $<BUILD_INTERFACE:${CMAKE_SOURCE_DIR}/include/RandLAPACK>
-    $<INSTALL_INTERFACE:include>
-    $<INSTALL_INTERFACE:include/RandLAPACK>
+    $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/../include>
+    $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/../include/RandLAPACK>
```
rileyjmurray, line 24: What files do we expect to be located in ``{CMAKE_CURRENT_SOURCE_DIR}/../``? Is this just robustness of the cmake build system?

src/CMakeLists.txt
```diff
@@ -5,25 +5,25 @@ set(RandLAPACK_cxx_srcs
...
 set(RandLAPACK_libs
-    RandLAPACK_RandBLAS
-    RandLAPACK_lapackpp
-    RandBLAS_blaspp
+    RandBLAS
+    lapackpp
+    blaspp
     RandBLAS_Random123
-    RandBLAS_OpenMP
+    OpenMP::OpenMP_CXX
```
rileyjmurray, line 15: What effect does removing the ``RandLAPACK_`` and ``RandBLAS_`` prefixes have here? It looks like ``RandBLAS``, ``lapackpp``, and ``blaspp`` will all have semantic interpretations within CMake. That is, CMake will associate the value of the build flag ``-Dlapackpp_DIR=...`` with ``lapackpp``. Is that correct? Maybe this is something for us to discuss on the call.

test/src/comps/test_determiter.cc
```diff
@@ -1,7 +1,7 @@
+#include <gtest/gtest.h>
```
rileyjmurray, line 1: Previously I got build errors when I included gtest before certain other libraries.  Are you moving it here for style reasons?

test/src/comps/test_precond.cc
```diff
@@ -72,9 +72,8 @@ class Test_rpc_svd_sjlt : public ::testing::Test
         );
         
         // check the result
-        double reldtol = RELDTOL;
         EXPECT_EQ(rank, n);
-        double *ignore;
+        double *ignore = nullptr;
```
rileyjmurray, line 76: Are you generally in favor of ``nullptr`` over ``NULL``, or do you determine which to use on a case-by-case basis?

rileyjmurray (commented):
I trust that these changes are good, but they raised several questions for me.

test/src/comps/test_precond.cc
burlen: `NULL` is a macro defined to be the integer 0. `nullptr`, is a key word added to the language in C++11, and is preferred because it has a pointer type, so that it doesn't automatically convert to an integer. This resolves ambiguity in template deduction and overload resolution between int and pointer types.

test/src/comps/test_determiter.cc
burlen: nope, to resolve build errors when it is included after lapackpp.

src/CMakeLists.txt
burlen: this is to support nesting projects. `CMAKE_SOURCE_DIR` is always the top level. When we are a git submodule it is the wrong path.

src/CMakeLists.txt
burlen: The prefixes disambiguated interface libraries that carried dependency information in the installs. I got rid of those because I think they are not needed. Instead we look for the dependency directly.

test/src/comps/test_precond.cc
burlen: That said I will not complain as long as your pointers are initialized to zero, any method will do the trick.

burlen:
the acceptable parts of these changes were merged with #12

PR #13, opened 2022-12-16
Fixes use from install

burlen:


rileyjmurray:
@burlen can you remind me what this PR is for?

burlen:
This fixes cmake related issues that arise when using RandLAPACK form another project.

burlen:
@rileyjmurray Much of the CMake work in `basic_svd_precond` needs to be in `main` we should figure out a strategy for getting it in.

PR #16, opened 2023-01-18
Update BLAS++ and LAPACK++ links.

TeachRaccooon:
Libraries migrated to github.


TeachRaccooon:
@burlen do you have an intuition behind why the build is failing here? All i added was updated links for blas++ and lapack++.

burlen:
I assume that once you rebase from main things will be fine. I'll see if there's a setting in github to make this clear. If things still fail after you do that, I'll take a closer look.

burlen:
yes, there is a setting, and it was not on. I enabled it and now github will tell you when you are out of date.

TeachRaccooon:
Looks like all checks are passing now, thanks! All in favor of merging?

PR #23, opened 2023-02-08
let the compiler deduce types

burlen:
- blas mixes templates for slower unoptimized fall back implementations and overloads for faster optimized implementations. specifying the template type in calls to blas functions tells the compiler you want to use the slower template implementation. For more info see: https://github.com/icl-utk-edu/blaspp/issues/31
- removes `using namespace blas` and `using namespace lapack`. this is largely for purposes of self documentation, but especially useful since blas and lapack define finctions that clash with the STL.
- adds include guards to our headers
- reorders includes. our files should be included first so that our configuration/definitions take precedence over other the other libraries'.
- use `include "x"` for our headers instead of `include <x>`. this is largely for the purposes of self documentation, but the two do have a different meaning.


include/RandLAPACK/drivers/rsvd.hh
```diff
@@ -1,9 +1,8 @@
-#ifndef BLAS_HH
-#include <blas.hh>
-#define BLAS_HH
-#endif
+#ifndef randlapack_drivers_rsvd_h
+#define randlapack_drivers_rsvd_h
 
 #include "../comps/qb.hh"
+#include "blaspp.h"
```
rileyjmurray, line 5: I'm confused about these two lines. The first line seems to imply that include statements with quotations are relative to the position of the current file. But if that's true, then the second include statement should be ``#include "../blaspp.h``. Is the current format really valid? If so, can you explain how these statements are being parsed?

Separately, I don't seriously want to support Windows, but is it okay to specify file paths using forward slashes when using C++ compilers on Windows machines?

rileyjmurray, line 5: I'm confused about these two lines. The first line seems to imply that include statements with quotations are relative to the position of the current file. But if that's true, then the second include statement should be ``#include "../blaspp.h``. Is the current format really valid? If so, can you explain how these statements are being parsed?

Separately, I don't seriously want to support Windows, but is it okay to specify file paths using forward slashes when using C++ compilers on Windows machines?

include/RandLAPACK/lapackpp.h
```diff
@@ -0,0 +1,12 @@
+#ifndef randlapack_lapackpp_h
+#define randlapack_lapackpp_h
+
+#define LAPACK_COMPLEX_CPP 1
```
rileyjmurray, line 4: What is this define statement for?

src/comps/orth.cc
```diff
@@ -1,10 +1,11 @@
+#include <RandLAPACK.hh>
+#include <RandBLAS.hh>
```
rileyjmurray, line 2: Two comments:

1. I think ``#include <RandLAPACK.hh>`` should be ``#include "RandLAPACK.hh"``.
2. It seems excessive to include all of RandLAPACK for this file. There is a good chance that ``#include "RandLAPACK/comps/orth.hh"`` would suffice.

src/comps/qb.cc
```diff
@@ -1,6 +1,7 @@
-#include <RandBLAS.hh>
-#include <lapack.hh>
-#include <RandLAPACK.hh>
+#include "RandLAPACK.hh"
```
rileyjmurray, line 1: Following my comment on ``src/comps/orth.cc``: I think this include statement of RandLAPACK can be changed to ``#include "RandLAPACK/comps/qb.hh"``.

src/comps/rf.cc
```diff
@@ -1,10 +1,11 @@
+#include "RandLAPACK.hh"
```
rileyjmurray, line 1: I think this can be ``#include "RandLAPACK/comps/rf.hh"``.

src/drivers/cholqrcp.cc
```diff
@@ -1,15 +1,13 @@
+#include "RandLAPACK.hh"
```
rileyjmurray, line 1: I think ``#include "RandLAPACK/drivers/cholqrcp.hh"`` will suffice.

src/drivers/rsvd.cc
```diff
@@ -1,13 +1,12 @@
+#include "RandLAPACK.hh"
```
rileyjmurray, line 1: I think ``#include "RandLAPACK/drivers/rsvd.hh"`` will suffice

test/src/comps/test_determiter.cc
```diff
@@ -1,8 +1,9 @@
-#include <blas.hh>
-#include <RandBLAS.hh>
-#include <RandLAPACK.hh>
-#include <gtest/gtest.h>
+#include "RandLAPACK.hh"
+#include "RandBLAS.hh"
```
rileyjmurray, line 2: Technically, isn't RandBLAS an external project, in the sense that it should be included with brackets when used in RandLAPACK?

test/src/comps/test_orth.cc
```diff
@@ -1,17 +1,14 @@
-/*
-TODO #1: Switch tuples to vectors.
-*/
+#include "RandBLAS.hh"
+#include "RandLAPACK.hh"
```
rileyjmurray, line 2: I think RandBLAS should be included with brackets.

test/src/comps/test_util.cc
```diff
@@ -1,18 +1,17 @@
+#include "RandLAPACK.hh"
+#include "RandBLAS.hh"
```
rileyjmurray, line 2: RandBLAS should probably be included with brackets.

test/src/drivers/test_cholqrcp.cc
```diff
@@ -1,10 +1,10 @@
-#include <gtest/gtest.h>
-#include <blas.hh>
-#include <lapack.hh>
-#include <RandBLAS.hh>
-#include <RandLAPACK.hh>
+#include "RandLAPACK.hh"
+#include "RandBLAS.hh"
```
rileyjmurray, line 2: Should probably include RandBLAS with brackets.

test/src/drivers/test_rsvd.cc
```diff
@@ -1,10 +1,10 @@
-#include <gtest/gtest.h>
-#include <blas.hh>
-#include <lapack.hh>
-#include <RandBLAS.hh>
-#include <RandLAPACK.hh>
+#include "RandBLAS.hh"
```
rileyjmurray, line 1: Should probably include RandBLAS with brackets

rileyjmurray (commented):
I left comments about includes in a few places. They are somewhat repetitive, so once we agree on what's going on then you can resolve all of them at once. We can discuss individual comments in comment-specific threads. But I do have higher-level questions about how includes work:

1. Is it the case that project-internal include statements processed within header files are relative to the file being processed? Like, we might have to specify a file that's one or two directories back?
2. Is it the case that project-internal include statements processed in ``.cc`` source files are processed relative to the project root?

(By "project-internal include statement," I mean ``#include "file_in_this_project.hh"``.)
The patterns I've seen suggest the answers are "yes," but I want to confirm.

include/RandLAPACK/drivers/rsvd.hh
burlen: quotes vs brackets change how the pre-processor searches for included files. with quotes it will look relative to the current directory first, but it will still fallback to the usual search if that fails. for that reason using quotes usually doesn't change much. 

Using quotes is a convention to signal to the reader that "this header is one of ours".  I think it's helpful do adopt such conventions it makes it easier to differentiate our code from our dependencies, for example `#include <blas.hh>` for the BLAS++ header vs `#include "blaspp.h"` for our header.

the first line works because the path is actually correct, the second line works because, the file is found in one of the paths we told the pre-processor to look in via a command line argument. I thought about change the first line, but wanted to limit scope.

 yes, microsoft's  c pre-processor can handle /, what a mess life would be if it didn't!

include/RandLAPACK/lapackpp.h
burlen: it tells LAPACK++ not to `#include <complex.h>`, the C header. Google test needs the C++ version in `#include <complex>`.

src/comps/orth.cc
burlen: agree on both points. I'll see if your suggestion works.

src/comps/qb.cc
burlen: OK, I'll give it  a try

test/src/comps/test_determiter.cc
burlen: lol, OK

src/comps/orth.cc
burlen: done

src/comps/qb.cc
burlen: done

test/src/drivers/test_cholqrcp.cc
burlen: done

test/src/comps/test_util.cc
burlen: done

test/src/comps/test_orth.cc
burlen: done

test/src/comps/test_determiter.cc
burlen: done

src/comps/rf.cc
burlen: done

burlen:
@rileyjmurray made the suggested changes. keep in mind that there are compiler's default paths, usually searched first, using "" makes it search the current dir first; and there are additional paths we specify on the command line, which are searched after.  did that answer your questions?

include/RandLAPACK/lapackpp.h
rileyjmurray: Gotcha. Thanks!

include/RandLAPACK/drivers/rsvd.hh
rileyjmurray: Gotcha. Makes sense. Thanks!

rileyjmurray (approved):

rileyjmurray:
> @rileyjmurray made the suggested changes. keep in mind that there are compiler's default paths, usually searched first, using "" makes it search the current dir first; and there are additional paths we specify on the command line, which are searched after. did that answer your questions?

Yup! PR looks good to me. Merge when ready.

PR #24, opened 2023-02-08
testing improvements

burlen:
- enable the address sanitizer in github actions
- run on macos

### DIFF
--- a/.github/workflows/core-linux.yaml
+++ b/.github/workflows/core-linux.yaml
@@ -65,17 +65,33 @@ jobs:
               `pwd`/../lapackpp
           make -j2 install
 
-      - name: build RandLAPACK
+      - name: build and test RandLAPACK (Release)
         run: |
           cd ..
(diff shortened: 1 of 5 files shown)


burlen:
this is ready to go. macos failure is from a test that's trying to use threads. openmp is not found in the macos environment, apple clang does not provide openmp. the absolute tolerance is too low,but really we could disable this test when openmp is not found.

rileyjmurray:
@burlen there are merge conflicts that prevent this from being merged into ``main`` as-is. Can you resolve those? Once they're taken care of then we can merge. (I have a plan for the failing tests. Let's let them continue to fail until I implement that fix.)

burlen:
@rileyjmurray just seeing this! Glad you have a plan for the tests. I'll resolve conflicts. I have a RandBLAS PR now with some improvements to the testing utils, but I was not able to resolve 1 of the test failures.

PR #29, opened 2023-03-23
Add random states

TeachRaccooon:
Substitutes the use of random seeds for the use of random states across RandLAPACK.


rileyjmurray:
Just want to point out: right now, you're always passing by value, as opposed to passing by reference. I think that's okay, but we really should consider passing by reference, and we should document our discussion on this point.

We'll need to revisit this once we update RandLAPACK to use the latest version of RandBLAS.

TeachRaccooon:
> Just want to point out: right now, you're always passing by value, as opposed to passing by reference. I think that's okay, but we really should consider passing by reference, and we should document our discussion on this point.
> 
> We'll need to revisit this once we update RandLAPACK to use the latest version of RandBLAS.

Yes, let me do that in the same PR when update the RandBLAS used.

PR #30, opened 2023-03-23
Rank estimation update

TeachRaccooon:
Added an option of choosing between naive and principled rank estimation in CQRRPT.
Added an option of checking the condition number of A^{pre} in CQRRPT.
Added a benchmark for monitoring approximation error and ran estimation by CQRRPT with the change of condition number of an input matrix.


RandLAPACK/comps/rl_util.hh
```diff
@@ -497,5 +497,36 @@ bool orthogonality_check(
...
+template <typename T>
+int64_t rank_search(
+    int64_t lo,
+    int64_t hi,
+    int64_t k,
+    int64_t n,
+    T norm_A,
+    T tau_trunc,
+    T* A_dat
+) {
```
rileyjmurray, line 510: Document this function.

RandLAPACK/drivers/rl_cqrrpt.hh
```diff
@@ -246,37 +256,70 @@ int CQRRPT<T>::CQRRPT1(
...
+        }
+
+        // find l2-norm of the full R
+        T norm_R = lapack::lange(Norm::Fro, n, n, R_dat, n);
+
+        T norm_R_sub = lapack::lange(Norm::Fro, 1, n, &R_dat[(n - 1) * n], 1);
+        // Chek if R is full column rank
+        if ((norm_R_sub > 0.000001 * norm_R))
+        {
+            k = n;
```
rileyjmurray, line 297: 1. There's a typo in ``Chek``.
2. Describe what you're testing here.

RandLAPACK/comps/rl_util.hh
TeachRaccooon: Done.

benchmark/CQRRPT_cond_orth_effect.cc
```diff
@@ -0,0 +1,163 @@
...
+    T* A_dat = A.data();
+    T* I_ref_dat = I_ref.data();
+
+    // Check orthogonality of Q
+    // Q' * Q  - I = 0
+    blas::gemm(Layout::ColMajor, Op::Trans, Op::NoTrans, n, n, m, 1.0, A_dat, m, A_dat, m, -1.0, I_ref_dat, n);
+    T norm_Q = lapack::lange(lapack::Norm::Fro, n, n, I_ref_dat, n);
+    printf("COND(A^{pre}): %21e\n", std::get<1>(mat_type));
+    printf("FRO NORM OF Q' * Q - I: %e\n\n", norm_Q);
+}
```
rileyjmurray, line 45: The orthogonality check requires that $\operatorname{rank}(A) = n$. But it looks like the problem setup code ensures $\operatorname{rank}(A) = \texttt{true}\\_\texttt{k}$, where it's possible to get $\texttt{true}\\_\texttt{k} < n$. Am I missing something here?

RandLAPACK/drivers/rl_cqrrpt.hh
TeachRaccooon: Done.

benchmark/CQRRPT_cond_orth_effect.cc
TeachRaccooon: No, you are corrent, true rank parameter should not be there.

RandLAPACK/comps/rl_util.hh
```diff
@@ -497,5 +497,43 @@ bool orthogonality_check(
...
+    T norm_R_sub = lapack::lange(Norm::Fro, n - k, n, &A_dat[k * n], n - k);
+
+    if(k == k + ((k - lo) / 2)) {
```
rileyjmurray, line 520: This seems like a weird way to write ``(k - lo) / 2 == 0``.

RandLAPACK/drivers/rl_cqrrpt.hh
rileyjmurray, line 1: You need to update the docstring for CQRRPT to that you mention ``naive_rank_estimate`` and any memory allocation you might do internally.

RandLAPACK/comps/rl_util.hh
TeachRaccooon: Lol yeah, you know me, i'm not looking for the easy ways.

RandLAPACK/drivers/rl_cqrrpt.hh
rileyjmurray, line 1: Question: why do you have ``CQRRPT1`` _and_ ``call`` functions for the ``CQRRPT`` class? It looks like you're just doing it to reduce code duplication in whether or not you print a status code. I suggest you remove the separate ``CQRRPT1`` function and move its functionality into ``call``. Then you can use a ``goto`` statement to jump to any status-printing and/or function-execution-cleanup code. Goto statements are generally discouraged (see https://www.learncpp.com/cpp-tutorial/goto-statements/), but it seems to me that using them in this kind of situation is far, far more readable than the nest one function within another paradigm you're currently using.

@burlen, what do you think?

TeachRaccooon: Not sure if i follow the goto logic, but i can just put the status updating statements inside of the call function itself. The only reason why this sort of logic is present is because we used to have multiple functions per class before, so call would be used to call several alternative algorithms. Would you prefer me dealing with this issue across RandLAPACK in this PR or in the next one?

TeachRaccooon: Done. Lmk if it looks ok

rileyjmurray: Prefer ``x == 0`` to ``!x`` if ``x`` is computed from a mathematical expression; improves readability.

rileyjmurray: Okay then we'll cross that ``goto`` bridge if/when we get to it.

RandLAPACK/drivers/rl_cqrrpt.hh
```diff
@@ -37,7 +37,19 @@ template <typename T>
...
+        ///
+        /// The algorithm allows for choosing the rank estimation schemeL either naively, through looking at the
+        /// diagonal entries of an R-factor from QRCP or via finding the smallest k such that ||A[k:, k:]||_F <= tau_trunk * ||A||_F.
+        /// This decision is controlled through 'naive_rank_estimate' parameter, which defaults to 1.
+        ///
+        /// The algorithm optionally times all of its subcomponents through a user-defined 'verbosity' parameter.
+        ///
+        /// The algorithm optionally computes a condition number of a preconditioned matrix A through a 'cond_check'
+        /// parameter, which defaults to 0. This requires extra n * (m + 1) * sizeof(T) bytes of space, which will be 
+        /// internally allocated by a utility routine.
```
rileyjmurray, line 52: Clarify: this is the $\ell_2$ condition number, right? If so, then would be good to mention that you compute this by computing all singular values with an appropriate LAPACK function.

rileyjmurray (changes requested):
Looks good other than minor comments. Resolve those and we can merge.

RandLAPACK/drivers/rl_cqrrpt.hh
```diff
@@ -37,7 +37,19 @@ template <typename T>
 class CQRRPT : public CQRRPTalg<T> {
     public:
 
-        // Constructor
+        /// The algorithm allows for choosing how QRCP is emplemented: either thropught LAPACK's GEQP3
+        /// or through a custom HQRRP function. This decision is controlled through 'no_hqrrp' parameter,
+        /// which defaults to 1.
+        ///
+        /// The algorithm allows for choosing the rank estimation schemeL either naively, through looking at the
```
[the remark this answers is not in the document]
TeachRaccooon: Fixed.

RandLAPACK/drivers/rl_cqrrpt.hh
TeachRaccooon: Fixed.

TeachRaccooon:
> Looks good other than minor comments. Resolve those and we can merge.

Fixed everything, please merge.

PR #31, opened 2023-05-03
Expansion

TeachRaccooon:
This branch was initially intened to simply expand some of the utility functionality in order to be able to test things in accordance to what was being discussed duing the CQRRPT project meetings. 
The branch ended up including a lot more than just that.

Below is the list of changes. 
We need to decide on what should actually be added into main branch.
Also, commits history is a mess, once again, because I was testing a lot of things externally and pulling changes from git. 

1. Updated names of data files in benchmarks. This ensures that we are in line in terms of naming with the newly-added [benchmarking data](https://github.com/BallisticLA/RandLAPACK-benchmarking-data) repo. 
2. Utility updates: 
    - Updated 'get_U' for capturing the upper-triangular portion of the matrix. We now have two version: one that overwrited the original data and one that saves an upper-triangular copy into the buffer.
    - Updated get_L routine to optionally overwrite the main diagonal with all 1's.
    - Added 'step,' 'spiked,' 'scaled' and 'breaking' matrix generators. 
    - Added a routine to check numerical rank of a given matrix.
    - Added a power iterations-based routine for finding l2 norm of a given matrix.
    - Added a function for looking for rank in an upper-triangular R-factor of QRCP linearly, per Oleg's suggestion.
    - Added a funtion for normalization of columns of a given matrix. 
    - Updated a routine for checking the condition number of a given matrix. 
3. CQRRPT
    - Added options to estimate rank naively or in a principled way, via l2 and Frobenius norm. If Norms are estimated in a principled way, user has further options to find rank using binary search or linearly.
    - Added a check for cholesky factorization failure.
    - Added a check for the condition number of a normalized A^{pre}.
4. Added an adaptive randomized eigendecomposition scheme that uses Nystrom preconditioning + a tester for it.


TeachRaccooon:
Some things to consider:

1. If we decide to keep different rank-estimation strategies in CQRRPT, we can potentially take up an object-oriented approach to those as well, passing in an appropriate rank estimation algorithm object into CQRRPT.
2. I don't think we need different rank search strategies. Binary search is definitely better in practice. 
3. Matrix generators API is a mess + some of the generators are not very useful. This needs to be cleaned up at some point. 
4. There is a lot of unnecessary space allocation throughout RandLAPACK (especially in tests/benchmarks). This can be fixed easily by reusing space, but the readability might suffer.

TeachRaccooon:
Closing this pull request, as it has been broken up into the following:
#32 #33 #34 #35.

PR #32, opened 2023-05-03
Naming and util update

TeachRaccooon:
1. Updated names of data files in benchmarks. This ensures that we are in line in terms of naming with the newly-added [benchmarking data repo](benchmarking data](https://github.com/BallisticLA/RandLAPACK-benchmarking-data).
2. Slightly modified existing utility functions + added an alternative version for get_U.


RandLAPACK/comps/rl_util.hh
```diff
@@ -82,21 +82,19 @@ void disp_diag(
...
-/// Overwrites the passed-in matrix.
-template <typename T> void get_L( int64_t m, int64_t n, std::vector<T>& L);
-
 template <typename T>
 void get_L(
     int64_t m,
     int64_t n,
-    std::vector<T>& L
+    std::vector<T>& L,
+    int overwrite_diagonal
```
rileyjmurray, line 90: C++ supports default arguments. So you can just set this to ``int overwrite_diagonal = true`` or ``int overwrite_diagonal = false`` depending on which you think is more common.

RandLAPACK/comps/rl_util.hh
```diff
@@ -129,6 +128,20 @@ void get_U(
     }
 }
 
+/// Eliminates the lower-triangular portion of A
```
rileyjmurray, line 131: I would say "zeros-out".

RandLAPACK/comps/rl_util.hh
rileyjmurray: You had a nice comment on lines 85 and 86. Why'd you remove those?

RandLAPACK/comps/rl_util.hh
```diff
@@ -355,14 +368,46 @@ void gen_exp_mat(
...
 
+/// Find the condition number of a given matrix A.
+template <typename T>
+T cond_num_check(
+    int64_t m,
+    int64_t n,
+    const std::vector<T>& A,
+    std::vector<T>& A_cpy,
+    std::vector<T>& s,
+    bool verbosity
```
rileyjmurray, line 379: replace ``verbosity'' with ``verbose``. The former word describes how verbose something is (which is not a binary) while the latter says whether or not something is verbose.

rileyjmurray (changes requested):
I have a few minor comments. Please address and then go ahead and merge without waiting for me. Then you can open a new PR.

RandLAPACK/comps/rl_util.hh
TeachRaccooon: I don't think i see a reason behind using a default here.

RandLAPACK/comps/rl_util.hh
TeachRaccooon: Fixed.

rileyjmurray:
The test is failing on macOS for numerical reasons in an old version of RandBLAS. The version of RandBLAS on main passes tests. All RandLAPACK code (the subject of this PR) passes tests. So this is fine to merge.

PR #33, opened 2023-05-04
Util add

TeachRaccooon:
Adding matrix generators, utility routines for finding l2 norm and normalizing columns of a matrix and tests for the latter two.

### DIFF
--- a/RandLAPACK/comps/rl_util.hh
+++ b/RandLAPACK/comps/rl_util.hh
@@ -273,10 +273,9 @@ void gen_mat(
 
 /// Generates matrix with the following singular values:
 /// sigma_i = 1 / (i + 1)^pow (first k * 0.2 sigmas = 1
-/// Can either be a diagonal matrix, or a full one.
+/// Output matrix is m by n of rank k.
 /// In later case, left and right singular vectors are randomly-generated
 /// and orthogonaized.
-/// Parameter 'k' signifies the numer of singular values in the generated matrix.
 /// Boolean parameter 'diag' signifies whether the matrix is to be
 /// generated as diagonal.
 /// Parameter 'cond' signfies the condition number of a generated matrix.
@@ -328,10 +327,9 @@ void gen_poly_mat(
 
 /// Generates matrix with the following singular values:
 /// sigma_i = e^((i + 1) * -pow) (first k * 0.2 sigmas = 1
-/// Can either be a diagonal matrix, or a full one.
+/// Output matrix is m by n of rank k.
 /// In later case, left and right singular vectors are randomly-generated
 /// and orthogonaized.
-/// Parameter 'k' signifies the numer of singular values in the generated matrix.
 /// Boolean parameter 'diag' signifies whether the matrix is to be
 /// generated as diagonal.
 /// Parameter 'cond' signfies the condition number of a generated matrix.
@@ -378,8 +376,8 @@ void gen_exp_mat(
     }
 }
 
-/// Generates matrix with a staircase spectrum with 4 steps
-/// Parameter 'k' signifies the numer of singular values in the generated matrix.
+/// Generates matrix with a staircase spectrum with 4 steps.
+/// Output matrix is m by n of rank k.
 /// Boolean parameter 'diag' signifies whether the matrix is to be
 /// generated as diagonal.
 /// Parameter 'cond' signfies the condition number of a generated matrix.
@@ -422,6 +420,7 @@ void gen_step_mat(
 }
 
 /// Generates a matrix with high coherence between the left singular vectors.
+/// Output matrix is m by n, full-rank.
 /// Such matrix would be difficult to sketch.
 /// Right singular vectors are sampled uniformly at random.
 template <typename T>
@@ -471,10 +470,12 @@ void gen_spiked_mat(
 
 /// Generates a numerically rank-deficient matrix.
 /// Added per Oleg's suggestion.
+/// Output matrix is m by n of some rank k < n.
 /// Generates a matrix of the form A = UV, where
 /// U is an m by n Gaussian matrix whose first row was scaled by a factor sigma, and that then 
-/// was orthonormalized with a Householder QR. The matrix V is the upper triangular part of an n × n 
-/// orthonormalized Gaussian matrix with modified diagonal entries to diag(V) *= [1, 10^-15, . . . , 10^-15, 10^-15]
+/// was orthonormalized with a Householder QR. 
+/// The matrix V is the upper triangular part of an n × n 
+/// orthonormalized Gaussian matrix with modified diagonal entries to diag(V) *= [1, 10^-15, . . . , 10^-15, 10^-15].
 template <typename T>
 void gen_oleg_adversarial_mat(
     int64_t& m,
@@ -522,8 +523,9 @@ void gen_oleg_adversarial_mat(
     blas::gemm(Layout::ColMajor, Op::NoTrans, Op::NoTrans, m, n, n, 1.0, U.data(), m, V.data(), n, 0.0, A.data(), m);
 }
 
-/// Per Oleg's suggestion, this matrix is supposed to break QB with Cholesky QR
-/// Parameter 'k' signifies the dimension of a sketching operator
+/// Per Oleg's suggestion, this matrix is supposed to break QB with Cholesky QR.
+/// Output matrix is m by n, full-rank.
+/// Parameter 'k' signifies the dimension of a sketching operator.
 /// Boolean parameter 'diag' signifies whether the matrix is to be
 /// generated as diagonal.
 /// Parameter 'cond' signfies the condition number of a generated matrix.
@@ -758,10 +760,9 @@ bool orthogonality_check(
 }
 
 /// Computes an L-2 norm of a given matrix using
-/// 10 steps of power iteration.
-/// Requires additional n * (n + 2) * sizeof(T) bytes of space.
+/// p steps of power iteration.
 template <typename T>
-T get_2_norm(
+T estimate_spectral_norm(
     int64_t m,
     int64_t n,
     T* A_dat,
@@ -839,8 +840,10 @@ void normc(
     T col_nrm = 0.0;
     for(int i = 0; i < n; ++i) {
         col_nrm = blas::nrm2(m, &A[m * i], 1);
-        for (int j = 0; j < m; ++j) {
-            A_norm[m * i + j] = A[m * i + j] / col_nrm;
+        if(col_nrm != 0) {
+            for (int j = 0; j < m; ++j) {
+                A_norm[m * i + j] = A[m * i + j] / col_nrm;
+            }
         }
     }
 }
--- a/RandLAPACK/comps/rl_util.hh
+++ b/RandLAPACK/comps/rl_util.hh
@@ -832,7 +832,7 @@ template <typename T>
 void normc(
     int64_t m,
     int64_t n,
-    std::vector<T>& A,
+    const std::vector<T>& A,
     std::vector<T>& A_norm
 ) {
     util::upsize(m * n, A_norm);
--- a/RandLAPACK/comps/rl_util.hh
+++ b/RandLAPACK/comps/rl_util.hh
@@ -826,29 +826,6 @@ int64_t rank_search_binary(
     return k;
 }
 
-/// Iterates the upper-triangular matrix A from end to beginning and 
-/// attempts to estimate smallest k such taht ||A[k:, k:]||_F <= tau_trunk * ||A||_2.
-/// Computes matrix Fro norms through vector norm updates.
-template <typename T>
-int64_t rank_search_linear(
-    int64_t n,
-    T norm_2_A,
-    T tau_trunc,
-    T* A_dat
-) {
-    T norm_A_work = 0.0;
-    T norm_A_row = 0.0;
-    for(int i = n - 1, j = 1; i > 0; --i, ++j) {
-        norm_A_row = blas::nrm2(j, A_dat + ((n + 1) * i), n);
-        // Add norm stably
-        norm_A_work = std::hypot(norm_A_work, norm_A_row);
-
-        if(norm_A_work > tau_trunc * norm_2_A)
-            return i + 1;
-    }
-    return 1;
-}
-
 /// Normalizes columns of a given matrix, writes the result into a buffer
 template <typename T>
 void normc(
--- a/RandLAPACK/comps/rl_util.hh
+++ b/RandLAPACK/comps/rl_util.hh
@@ -276,6 +276,10 @@ void gen_mat(
 /// Can either be a diagonal matrix, or a full one.
 /// In later case, left and right singular vectors are randomly-generated
 /// and orthogonaized.
+/// Parameter 'k' signifies the numer of singular values in the generated matrix.
+/// Boolean parameter 'diag' signifies whether the matrix is to be
+/// generated as diagonal.
+/// Parameter 'cond' signfies the condition number of a generated matrix.
 template <typename T>
 void gen_poly_mat(
     int64_t& m,
@@ -327,6 +331,10 @@ void gen_poly_mat(
 /// Can either be a diagonal matrix, or a full one.
 /// In later case, left and right singular vectors are randomly-generated
 /// and orthogonaized.
+/// Parameter 'k' signifies the numer of singular values in the generated matrix.
+/// Boolean parameter 'diag' signifies whether the matrix is to be
+/// generated as diagonal.
+/// Parameter 'cond' signfies the condition number of a generated matrix.
 template <typename T>
 void gen_exp_mat(
     int64_t& m,
@@ -371,6 +379,10 @@ void gen_exp_mat(
 }
 
 /// Generates matrix with a staircase spectrum with 4 steps
+/// Parameter 'k' signifies the numer of singular values in the generated matrix.
+/// Boolean parameter 'diag' signifies whether the matrix is to be
+/// generated as diagonal.
+/// Parameter 'cond' signfies the condition number of a generated matrix.
 template <typename T>
 void gen_step_mat(
     int64_t& m,
@@ -411,41 +423,24 @@ void gen_step_mat(
 
 /// Generates a matrix with high coherence between the left singular vectors.
 /// Such matrix would be difficult to sketch.
+/// Right singular vectors are sampled uniformly at random.
 template <typename T>
 void gen_spiked_mat(
     int64_t& m,
     int64_t& n,
     std::vector<T>& A,
+    T spike_scale,
     RandBLAS::base::RNGState<r123::Philox4x32> state
 ) {
-
-    T* A_dat = A.data();
+    T* A_dat = upsize(m * n, A);
 
     int64_t num_rows_sampled = n / 2;
 
+    /// We use the indices of nonzeros for columns of a num_rows_sampled by m LASO with one nonzero per row.
     RandBLAS::sparse::SparseDist DS = {RandBLAS::sparse::SparseDistName::LASO, num_rows_sampled, m, 1};
     RandBLAS::sparse::SparseSkOp<T> S(DS, state, NULL, NULL, NULL);
     RandBLAS::sparse::fill_sparse(S);
 
-    // Fill A with stacked identities
-    int start = 0;
-    while(start + n <= m){
-        for(int j = 0; j < n; ++j) {
-            A_dat[start + (m * j) + j] = 1.0;
-        }
-        start += n;
-    }
-
-    // Scale randomly sampled rows
-    start = 0;
-    while (start + m <= m * n) {
-        for(int i = 0; i < num_rows_sampled; ++i) {
-            A_dat[start + (S.cols)[i] - 1] *= m;
-        }
-        start += m;
-    }
-
-    std::vector<T> A_hat(m * n, 0.0);
     std::vector<T> V(n * n, 0.0);
     std::vector<T> tau(n, 0.0);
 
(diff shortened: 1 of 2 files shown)


RandLAPACK/comps/rl_util.hh
```diff
@@ -370,6 +370,189 @@ void gen_exp_mat(
...
+template <typename T>
+void gen_step_mat(
+    int64_t& m,
+    int64_t& n,
+    std::vector<T>& A,
+    int64_t k,
+    T cond,
+    bool diagon,
+    RandBLAS::base::RNGState<r123::Philox4x32> state
+) {
```
rileyjmurray, line 383: Say what k does, how cond is used, and what ``diagon`` does.

RandLAPACK/comps/rl_util.hh
```diff
@@ -370,6 +370,189 @@ void gen_exp_mat(
...
+/// Generates matrix with a staircase spectrum with 4 steps
+template <typename T>
+void gen_step_mat(
+    int64_t& m,
+    int64_t& n,
+    std::vector<T>& A,
+    int64_t k,
+    T cond,
+    bool diagon,
+    RandBLAS::base::RNGState<r123::Philox4x32> state
```
rileyjmurray, line 382: Don't template with a specific RNGState type. Start by having ``template <typename T, typename RNG>`` before the function name. Then, for the ``state`` argument, declare it as type ``RandBLAS::base::RNGState<RNG>`` rather than ``RandBLAS::base::RNGState<r123::Philox4x32>``.

Please make this change throughout ``rl_util.hh`` as appropriate. Once you do this, please make sure you don't include Random123 in this file.

RandLAPACK/comps/rl_util.hh
```diff
@@ -370,6 +370,189 @@ void gen_exp_mat(
...
+) {
+
+    T* A_dat = A.data();
+
+    int64_t num_rows_sampled = n / 2;
+
+    RandBLAS::sparse::SparseDist DS = {RandBLAS::sparse::SparseDistName::LASO, num_rows_sampled, m, 1};
+    RandBLAS::sparse::SparseSkOp<T> S(DS, state, NULL, NULL, NULL);
+    RandBLAS::sparse::fill_sparse(S);
```
rileyjmurray, line 428: Add a comment before these lines, explaining that we use the indices of nonzeros for columns of a ``num_rows_sampled`` by ``m`` LASO with one nonzero per row.

RandLAPACK/comps/rl_util.hh
```diff
@@ -370,6 +370,189 @@ void gen_exp_mat(
...
+/// Such matrix would be difficult to sketch.
+template <typename T>
+void gen_spiked_mat(
+    int64_t& m,
+    int64_t& n,
+    std::vector<T>& A,
+    RandBLAS::base::RNGState<r123::Philox4x32> state
+) {
+
+    T* A_dat = A.data();
```
rileyjmurray, line 422: In your other functions, you automatically upsize ``std::vector`` before trying to fill it. You should probably do that here as well.

RandLAPACK/comps/rl_util.hh
```diff
@@ -370,6 +370,189 @@ void gen_exp_mat(
...
+            A_dat[start + (m * j) + j] = 1.0;
+        }
+        start += n;
+    }
+
+    // Scale randomly sampled rows
+    start = 0;
+    while (start + m <= m * n) {
+        for(int i = 0; i < num_rows_sampled; ++i) {
+            A_dat[start + (S.cols)[i] - 1] *= m;
```
rileyjmurray, line 443: Please have this function accept a proportionality parameter ``prop_n`` so that ``num_rows_sampled = prop_n * n`` (right now, you've hard-coded ``prop_n = 0.5``).

Please also have it accept a parameter ``spike_scale`` that controls the scale of the spiked rows (right now, you have a hard-coded factor of ``m``).

RandLAPACK/comps/rl_util.hh
```diff
@@ -370,6 +370,189 @@ void gen_exp_mat(
...
+    lapack::ungqr(n, n, n, V.data(), n, tau.data());
+
+    std::copy(A.data(), A.data() + (m * n), A_hat.data());
+    blas::gemm(Layout::ColMajor, Op::NoTrans, Op::Trans, m, n, n, 1.0, A_hat.data(), m, V.data(), n, 0.0, A.data(), m);
+}
+
+/// Generates a numerically rank-deficient matrix.
+/// Added per Oleg's suggestion.
+template <typename T>
+void gen_scaled_mat(
```
rileyjmurray, line 465: I suggest we use a more informative function name.

Perhaps ``gen_oleg_adversarial_mat``. (This would not be a great name but it would at least remind us to revisit the naming question later on.)

RandLAPACK/comps/rl_util.hh
```diff
@@ -370,6 +370,189 @@ void gen_exp_mat(
...
+    T* V_dat = V.data();
+    for(int i = 11; i < n; ++i)
+        V_dat[n * i + i] *= scaling_factor_V;
+
+    blas::gemm(Layout::ColMajor, Op::NoTrans, Op::NoTrans, m, n, n, 1.0, U.data(), m, V.data(), n, 0.0, A.data(), m);
+}
+
+/// Per Oleg's suggestion, this matrix is supposed to break QB with Cholesky QR
+template <typename T>
+void gen_break_mat(
```
rileyjmurray, line 513: How about ``gen_bad_cholqr_mat``

RandLAPACK/comps/rl_util.hh
```diff
@@ -370,6 +370,189 @@ void gen_exp_mat(
...
+    int64_t& m,
+    int64_t& n,
+    std::vector<T>& A,
+    int64_t k, // this is equal to the dimension of a sketching operator
+    T cond,
+    bool diagon,
```
rileyjmurray, line 519: Document the meaning of this parameter

RandLAPACK/comps/rl_util.hh
```diff
@@ -402,6 +585,24 @@ T cond_num_check(
...
+    int64_t m,
+    int64_t n,
+    const std::vector<T>& A
+) {
+    std::vector<T> A_pre_cpy;
+    std::vector<T> s;
+    RandLAPACK::util::cond_num_check(m, n, A, A_pre_cpy, s, false);
+
+    for(int i = 0; i < n; ++i) {
+        if (s[i] / s[0] <= 5 * std::numeric_limits<double>::epsilon())
```
rileyjmurray, line 600: Should have ``std::numeric_limits<T>::epsilon()``. You've currently hardcoded ``T = double``.

RandLAPACK/comps/rl_util.hh
```diff
@@ -477,6 +678,30 @@ void gen_mat_type(
...
+        case 7: {
+                // Generating matrix with a staircase-like spectrum
+                RandLAPACK::util::gen_step_mat(m, n, A, k, std::get<1>(type), std::get<2>(type), state);
+            }    
+            break;
+        case 8: {
+                // This matrix may be numerically rank deficient
+                RandLAPACK::util::gen_spiked_mat(m, n, A, state);
+                if(std::get<2>(type))
+                    k = rank_check(m, n, A);
```
rileyjmurray, line 690: ``rank_check`` doesn't raise any errors. I also don't see where you use ``k`` later on. So it's not clear to me what the rank check accomplishes here.

RandLAPACK/comps/rl_util.hh
```diff
@@ -513,6 +738,42 @@ bool orthogonality_check(
     return false;
 }
 
+/// Computes an L-2 norm of a given matrix using
```
rileyjmurray, line 741: GitHub won't let me comment on Line 725 in the ``orthogonality_check`` function, so I comment on it here.

Don't use ``gemm`` to form a gram matrix. Use ``syrk`` and then copy the data from the upper to the lower triangle (or the other way around).

RandLAPACK/comps/rl_util.hh
```diff
@@ -513,6 +738,42 @@ bool orthogonality_check(
...
+    for(int i = 1; i < n; ++i)
+        blas::copy(n - i, ATA.data() + i + ((i-1) * n), 1, ATA.data() + (i - 1) + (i * n), n);
+
+    T prev_norm_inv = 1.0;
+    for(int i = 0; i < p; ++i)
+    {
+        gemv(Layout::ColMajor, Op::NoTrans, n, n, prev_norm_inv, ATA.data(), n, buf.data(), 1, 0.0, buf1.data(), 1);
+        buf = buf1;
+        prev_norm_inv = 1 / blas::nrm2(n, buf.data(), 1);
+    }
```
rileyjmurray, line 772: My dude, I'm gonna put a hard no on this implementation.

Don't form the gram matrix explicitly. Apply it to a vector by first making a ``gemv`` call for ``u = A*v`` and then making a ``gemv`` call for ``w = A' * u``. The end result is ``u = A'Av``.

RandLAPACK/comps/rl_util.hh
```diff
@@ -513,6 +738,42 @@ bool orthogonality_check(
...
+    // Compute A^T*A
+    blas::syrk(Layout::ColMajor, Uplo::Lower, Op::Trans, n, m, 1.0, A_dat, m, 0.0, ATA.data(), n);
+    // Fill the upper triangular part of A^T*A
+    for(int i = 1; i < n; ++i)
+        blas::copy(n - i, ATA.data() + i + ((i-1) * n), 1, ATA.data() + (i - 1) + (i * n), n);
+
+    T prev_norm_inv = 1.0;
+    for(int i = 0; i < p; ++i)
+    {
+        gemv(Layout::ColMajor, Op::NoTrans, n, n, prev_norm_inv, ATA.data(), n, buf.data(), 1, 0.0, buf1.data(), 1);
```
rileyjmurray, line 769: Less important note: don't use ``gemv`` if you know your matrix is symmetric or Hermitian. You should use ``symv`` or ``hemv`` in these settings.

rileyjmurray: This comment isn't super relevant though because I want you to delete this code and not form the gram matrix explicitly.

RandLAPACK/comps/rl_util.hh
```diff
@@ -552,5 +813,47 @@ int64_t rank_search_binary(
...
+    T norm_A_row = 0.0;
+    for(int i = n - 1, j = 1; i > 0; --i, ++j) {
+        norm_A_row = blas::nrm2(j, A_dat + ((n + 1) * i), n);
+        // Add norm stably
+        norm_A_work = std::hypot(norm_A_work, norm_A_row);
+
+        if(norm_A_work > tau_trunc * norm_2_A)
+            return i + 1;
+    }
+    return 1;
```
rileyjmurray, line 836: What happens when this function is applied to the zero matrix?

RandLAPACK/comps/rl_util.hh
```diff
@@ -552,5 +813,47 @@ int64_t rank_search_binary(
...
+    std::vector<T>& A_norm
+) {
+    util::upsize(m * n, A_norm);
+
+    T col_nrm = 0.0;
+    for(int i = 0; i < n; ++i) {
+        col_nrm = blas::nrm2(m, &A[m * i], 1);
+        for (int j = 0; j < m; ++j) {
+            A_norm[m * i + j] = A[m * i + j] / col_nrm;
+        }
```
rileyjmurray, line 854: If ``A`` is column-major then you can just set ``inv_col_nrm = 1/col_nrm`` and then use ``blas::scal``.

rileyjmurray (changes requested):
I've left many comments. I didn't look at your tests but am about to go to lunch. Will give feedback on tests in the next round.

RandLAPACK/comps/rl_util.hh
rileyjmurray: @TeachRaccooon I'm calling attention to this comment. You'll need fixes throughout this file.

RandLAPACK/comps/rl_util.hh
TeachRaccooon: Fixed.

RandLAPACK/comps/rl_util.hh
TeachRaccooon: Done.

RandLAPACK/comps/rl_util.hh
TeachRaccooon: Done.

RandLAPACK/comps/rl_util.hh
```diff
@@ -370,6 +370,189 @@ void gen_exp_mat(
...
+
+    T* A_dat = A.data();
+
+    int64_t num_rows_sampled = n / 2;
+
+    RandBLAS::sparse::SparseDist DS = {RandBLAS::sparse::SparseDistName::LASO, num_rows_sampled, m, 1};
+    RandBLAS::sparse::SparseSkOp<T> S(DS, state, NULL, NULL, NULL);
+    RandBLAS::sparse::fill_sparse(S);
+
+    // Fill A with stacked identities
```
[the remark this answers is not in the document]
TeachRaccooon: Done.

RandLAPACK/comps/rl_util.hh
```diff
@@ -370,6 +370,189 @@ void gen_exp_mat(
...
+            A.resize(k * k);
+        }
+        lapack::lacpy(MatrixType::General, k, k, S.data(), k, A.data(), k);
+    } else {
+        gen_mat(m, n, A, k, S, state);
+    }
+}
+
+/// Generates a matrix with high coherence between the left singular vectors.
+/// Such matrix would be difficult to sketch.
```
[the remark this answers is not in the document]
TeachRaccooon: Done.

RandLAPACK/comps/rl_util.hh
```diff
@@ -370,6 +370,189 @@ void gen_exp_mat(
...
+    // Scale randomly sampled rows
+    start = 0;
+    while (start + m <= m * n) {
+        for(int i = 0; i < num_rows_sampled; ++i) {
+            A_dat[start + (S.cols)[i] - 1] *= m;
+        }
+        start += m;
+    }
+
+    std::vector<T> A_hat(m * n, 0.0);
```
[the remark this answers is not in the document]
TeachRaccooon: Ture. Done.

RandLAPACK/comps/rl_util.hh
```diff
@@ -370,6 +370,189 @@ void gen_exp_mat(
...
+    // Fill A with stacked identities
+    int start = 0;
+    while(start + n <= m){
+        for(int j = 0; j < n; ++j) {
+            A_dat[start + (m * j) + j] = 1.0;
+        }
+        start += n;
+    }
+
+    // Scale randomly sampled rows
```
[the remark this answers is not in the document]
TeachRaccooon: I'm not sure what is the metric of it working here.

RandLAPACK/comps/rl_util.hh
```diff
@@ -370,6 +370,189 @@ void gen_exp_mat(
...
+
+    lapack::geqrf(n, n, V.data(), n, tau.data());
+    lapack::ungqr(n, n, n, V.data(), n, tau.data());
+
+    std::copy(A.data(), A.data() + (m * n), A_hat.data());
+    blas::gemm(Layout::ColMajor, Op::NoTrans, Op::Trans, m, n, n, 1.0, A_hat.data(), m, V.data(), n, 0.0, A.data(), m);
+}
+
+/// Generates a numerically rank-deficient matrix.
+/// Added per Oleg's suggestion.
```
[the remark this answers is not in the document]
TeachRaccooon: Done.

RandLAPACK/comps/rl_util.hh
TeachRaccooon: Sure, whatever works.

RandLAPACK/comps/rl_util.hh
TeachRaccooon: Done.

RandLAPACK/comps/rl_util.hh
TeachRaccooon: Done.

RandLAPACK/comps/rl_util.hh
TeachRaccooon: This function has a very precise use case and will not be applied to a zero matrix.

RandLAPACK/comps/rl_util.hh
TeachRaccooon: Will that actually be faster?

RandLAPACK/comps/rl_util.hh
TeachRaccooon: Yeah, that's definitely better.
Sorry bout that.
Fixed.

RandLAPACK/comps/rl_util.hh
TeachRaccooon: Understood!

RandLAPACK/comps/rl_util.hh
TeachRaccooon: Updated it.

RandLAPACK/comps/rl_util.hh
TeachRaccooon: Per slack concersation: wil fix later.

RandLAPACK/comps/rl_util.hh
TeachRaccooon: Fixed the latter. We should discuss the former.

RandLAPACK/comps/rl_util.hh
rileyjmurray: No, but it will be more readable.

rileyjmurray: *More readable for people who are used to reading linear algebra code. There's a "reasonable" expectation that if a BLAS function exists for some operation, then you're going to use it. Framed another way, if people see a for loop that operates on the entries of a matrix they will assume that the purpose of the for loop is something that can't be done with one BLAS call.

RandLAPACK/comps/rl_util.hh
rileyjmurray: Either add that requirement to the documentation or include the special casing to catch the zero matrix. Either is fine with me.

TeachRaccooon: Also, i kind of feel like we can get rid of linear search. It inly complicates things.

rileyjmurray: I'm fine with removing linear search

RandLAPACK/comps/rl_util.hh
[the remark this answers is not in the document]
rileyjmurray: Actually, looking at it again I see that it just scales the same row twice. That's the desired behavior, so I'll mark this as resolved.

RandLAPACK/comps/rl_util.hh
rileyjmurray: Ah, didn't see that k was provided by the user here.

RandLAPACK/comps/rl_util.hh
rileyjmurray: @TeachRaccooon please make sure binary search works with the zero matrix.

test/comps/test_util.cc
```diff
@@ -28,7 +28,50 @@ class TestUtil : public ::testing::Test
 
     template <typename T>
     static void 
-    test_1()
-    {
+    test_2_norm(int64_t m, int64_t n, std::tuple<int, T, bool> mat_type, RandBLAS::base::RNGState<r123::Philox4x32> state) {
```
rileyjmurray, line 31: ``test_matrix_2_norm`` or ``test_spectral_norm``

rileyjmurray: Please rename the utility function itself to ``matrix_2_norm`` or ``spectral_norm``.

test/comps/test_util.cc
```diff
@@ -28,7 +28,50 @@ class TestUtil : public ::testing::Test
...
+        T norm = blas::nrm2(m, A_norm.data(), 1);
+
+        ASSERT_NEAR(norm, std::sqrt(m), std::pow(std::numeric_limits<double>::epsilon(), 0.75));
     }
 };
+
+TEST_F(TestUtil, test_2_norm) {
+    auto state = RandBLAS::base::RNGState(0, 0);
+    test_2_norm<double>(1000, 100, std::make_tuple(0, 2, false), state);
+    test_2_norm<double>(1000, 100, std::make_tuple(9, std::pow(10, 15), false), state);
```
rileyjmurray, line 71: Please also test single precision (you'll want to make sure the test matrix has max singular value <= 1e8).

test/comps/test_util.cc
```diff
@@ -28,7 +28,50 @@ class TestUtil : public ::testing::Test
...
+}
+
+TEST_F(TestUtil, test_normc) {
+    auto state = RandBLAS::base::RNGState(0, 0);
+    test_normc<double>(1000);
+}
```
rileyjmurray, line 77: Test what happens when a matrix has some column(s) equal to zero.

RandLAPACK/comps/rl_util.hh
```diff
@@ -513,6 +757,36 @@ bool orthogonality_check(
     return false;
 }
 
+/// Computes an L-2 norm of a given matrix using
+/// 10 steps of power iteration.
+/// Requires additional n * (n + 2) * sizeof(T) bytes of space.
```
rileyjmurray, line 762: I don't think this space requirement is true anymore.

rileyjmurray (changes requested):

test/comps/test_util.cc
rileyjmurray: Or maybe ``estimate_matrix_2_norm`` or ``estimate_spectral_norm``

RandLAPACK/comps/rl_util.hh
TeachRaccooon: Fixed. It is also not 10 steps anymore.

test/comps/test_util.cc
TeachRaccooon: Done. went with estimate_spectral_norm.

RandLAPACK/comps/rl_util.hh
```diff
@@ -276,6 +276,10 @@ void gen_mat(
 /// Can either be a diagonal matrix, or a full one.
 /// In later case, left and right singular vectors are randomly-generated
 /// and orthogonaized.
+/// Parameter 'k' signifies the numer of singular values in the generated matrix.
+/// Boolean parameter 'diag' signifies whether the matrix is to be
+/// generated as diagonal.
+/// Parameter 'cond' signfies the condition number of a generated matrix.
```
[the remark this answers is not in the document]
TeachRaccooon: Done.

test/comps/test_util.cc
TeachRaccooon: Good call, there was actually an issue here. Fixed.

RandLAPACK/comps/rl_util.hh
TeachRaccooon: Well, we're iterating through all entries in a given column-vector anyway. Why not scale those immediately instead of separately scaling the column vector after we've already assigned all values to it?

RandLAPACK/comps/rl_util.hh
TeachRaccooon: Done.

RandLAPACK/comps/rl_util.hh
rileyjmurray: Per your comment on Slack

> The problem is that the API for the matrix generation now only has 3 parameters that the user can specify: mat type, an arbitrary parameter and whether they want a diagonal matrix. We're already using an arbitrary one for the scaling factor. There are ways to make your idea work, but that requires an API change, which means fixing it all around RandLAPACK.

I'm fine leaving the latter unaddressed.

RandLAPACK/comps/rl_util.hh
rileyjmurray: @TeachRaccooon I actually missed the fact that ``A_norm`` was separate from ``A``.  It's obvious now.

You could do away with the second for-loop if ``A_norm`` started as all zeros. In that case, you could ``axpy`` each column of ``A`` into ``A_norm`` with a scaling factor ``1/col_nrm``. But since you don't have an assurance that ``A_norm`` is all zeros, your current implementation is fine.

One request: please go through these utility functions and mark values as ``const`` if the function isn't supposed to change them. In this case, I think you'd want the third argument to the function to be ``const std::vector<T>& A``. Try that out, and if it works, then please apply changes throughout ``rl_util.hh``.

TeachRaccooon: Done.
I actually did that before.
This function was the only place that needed change.

rileyjmurray:
@TeachRaccooon GitHub is losing track of my comments in the PR review feature. But you've addressed pretty much everything. Please do one last check for arguments to functions that should be ``const``. Look at pointer arguments and std::vector arguments. We can merge once you've done that last pass.

rileyjmurray (approved):
Looks good!

PR #35, opened 2023-05-05
Adding adaptive REVD2

TeachRaccooon:
Adding an adaptive REVD2 scheme + unit tests.

### DIFF
--- a/RandLAPACK/drivers/rl_revd2.hh
+++ b/RandLAPACK/drivers/rl_revd2.hh
@@ -124,47 +124,47 @@ T power_error_est(
     int64_t m,
     int64_t k,
     int p,
-    T* vector_buf_dat,
-    T* V_dat,
+    T* vector_buf,
+    T* V,
     blas::Uplo uplo,
-    T* A_dat,
-    T* Mat_buf_dat,
-    T* eigvals_dat
+    T* A,
+    T* Mat_buf,
+    T* eigvals
 ) {
     T err = 0;
         
     for(int i = 0; i < p; ++i) {
-        T g_norm = blas::nrm2(m, vector_buf_dat, 1);
+        T g_norm = blas::nrm2(m, vector_buf, 1);
         // Compute g = g / ||g|| - we need this because dot product does not take in an alpha
-        std::for_each(vector_buf_dat, vector_buf_dat + m, [g_norm](T &entry) { entry /= g_norm;});
+        blas::scal(m, 1 / g_norm, vector_buf, 1);
 
         // Compute V'*g / ||g||
         // Using the second column of Omega as a buffer for matrix-vector product
-        gemv(Layout::ColMajor, Op::Trans, m, k, 1.0, V_dat, m, vector_buf_dat, 1, 0.0, &vector_buf_dat[m], 1);
+        gemv(Layout::ColMajor, Op::Trans, m, k, 1.0, V, m, vector_buf, 1, 0.0, &vector_buf[m], 1);
 
 
         // Compute V*E, eigvals diag
         // Using Y as a buffer for V*E
         for (int i = 0, j = 0; i < m * k; ++i) {
-            Mat_buf_dat[i] = V_dat[i] * eigvals_dat[j];
+            Mat_buf[i] = V[i] * eigvals[j];
             if((i + 1) % m == 0 && i != 0)
                 ++j;
         }
 
         // Compute V*E*V'*g / ||g||
         // Using the third column of Omega as a buffer for matrix-vector product
-        gemv(Layout::ColMajor, Op::NoTrans, m, k, 1.0, Mat_buf_dat, m, &vector_buf_dat[m], 1, 0.0, &vector_buf_dat[2 * m], 1);
+        gemv(Layout::ColMajor, Op::NoTrans, m, k, 1.0, Mat_buf, m, &vector_buf[m], 1, 0.0, &vector_buf[2 * m], 1);
         // Compute A*g / ||g||
         // Using the forth column of Omega as a buffer for matrix-vector product
-        symv(Layout::ColMajor, uplo, m, 1.0, A_dat, m, vector_buf_dat, 1, 0.0, &vector_buf_dat[3 * m], 1);
+        symv(Layout::ColMajor, uplo, m, 1.0, A, m, vector_buf, 1, 0.0, &vector_buf[3 * m], 1);
 
         // Compute A*g / ||g|| - V*E*V'*g / ||g||
         // Result is stored in the 4th column of Omega
-        blas::axpy(m, -1.0, &vector_buf_dat[2 * m], 1, &vector_buf_dat[3 * m], 1);
+        blas::axpy(m, -1.0, &vector_buf[2 * m], 1, &vector_buf[3 * m], 1);
         // Compute (g / ||g||)' * (A*g / ||g|| - V*E*V'*g / ||g||) - this is our measure for the error
-        err = blas::dot(m, vector_buf_dat, 1, &vector_buf_dat[3 * m], 1);	
+        err = blas::dot(m, vector_buf, 1, &vector_buf[3 * m], 1);	
         // v_0 <- v
-        std::copy(&vector_buf_dat[3 * m], &vector_buf_dat[4 * m], vector_buf_dat);
+        std::copy(&vector_buf[3 * m], &vector_buf[4 * m], vector_buf);
     }
     return err;
 }
--- a/RandLAPACK/drivers/rl_revd2.hh
+++ b/RandLAPACK/drivers/rl_revd2.hh
@@ -1,0 +1,240 @@
+#ifndef randlapack_drivers_revd2_h
+#define randlapack_drivers_revd2_h
+
+#include "rl_qb.hh"
+#include "rl_blaspp.hh"
+#include "rl_lapackpp.hh"
+#include "rl_util.hh"
+
+#include <RandBLAS.hh>
+#include <cstdint>
+#include <vector>
+
+namespace RandLAPACK {
+
+template <typename T>
+class REVD2alg {
+    public:
+
+        virtual ~REVD2alg() {}
+
+        virtual int call(
+            int64_t m,
+            std::vector<T>& A,
+            int64_t& k,
+            std::vector<T>& V,
+            std::vector<T>& E
+        ) = 0;
+};
+
+template <typename T>
+class REVD2 : public REVD2alg<T> {
+    public:
+
+        // Constructor
+        REVD2(
+            RandLAPACK::RangeFinder<T>& rf_obj,
+            T tolerance,
+            int power_iters,
+            RandBLAS::base::RNGState<r123::Philox4x32> s,
+            bool verb
+        ) : RF_Obj(rf_obj) {
+            tol = tolerance;
+            p = power_iters;
+            state = s;
+            verbosity = verb;
+        }
+
+        /// Computes a rank-k approximation to an EVD of a symmetric positive definite matrix:
+        ///     A= V diag(E) V*,
+        /// where V is a matrix of eigenvectors and E is a vector of eigenvalues.
+        /// Does so adaptively. 
+        /// If the tolerance is not met, increases the rank estimation parameter by 2.
+        /// Tolerance is the maximum of user-specified tol times 5 and computed 'nu' times 5.
+        /// This is motivated by the fact that the approx error will never be smaller than nu.
+        /// This code is identical to ALgorithm E2 from https://arxiv.org/pdf/2110.02820.pdf.
+        /// Uses RangeFinder for constructing a sketching operator.
+        /// Has a lot of potential in terms of storage space optimization, 
+        /// which, however, will affect readability.
+        ///
+        /// @param[in] m
+        ///     The number of rows in the matrix A.
+        ///
+        /// @param[in] n
+        ///     The number of columns in the matrix A.
+        ///
+        /// @param[in] A
+        ///     The m-by-m matrix A, stored in a column-major format.
+        ///     Must be SPD.
+        ///
+        /// @param[in] k
+        ///     Column dimension of a sketch, k <= n.
+        ///
+        /// @param[in] V
+        ///     On entry, is empty and may not have any space allocated for it.
+        ///
+        /// @param[in] E
+        ///     On entry, is empty and may not have any space allocated for it.
+        ///
+        /// @param[out] V
+        ///     Stores m-by-k matrix matrix of eigenvectors.
+        ///
+        /// @param[out] E
+        ///     Stores k eigenvalues.
+        ///
+        /// @return = 0: successful exit
+        ///
+
+        int call(
+            int64_t m,
+            std::vector<T>& A,
+            int64_t& k,
+            std::vector<T>& V,
+            std::vector<T>& E
+        ) override;
+
+    public:
+        RandLAPACK::RangeFinder<T>& RF_Obj;
+        RandBLAS::base::RNGState<r123::Philox4x32> state;
+        T tol;
+        int p;
+        bool verbosity;
+
+        std::vector<T> Y;
+        std::vector<T> Omega;
+
+        std::vector<T> R;
+        std::vector<T> S;
+
+        std::vector<T> A_cpy;
+};
+
+// -----------------------------------------------------------------------------
+template <typename T>
+int REVD2<T>::call(
+        int64_t m, // m = n
+        std::vector<T>& A,
+        int64_t& k,
+        std::vector<T>& V,
+        std::vector<T>& E
+){
+    T err = 0;
+    while(1) {
+        T* A_dat = A.data();
+        T* V_dat = util::upsize(m * k, V);
+        util::upsize(k, E);
+        T* Y_dat = util::upsize(m * k, this->Y);
+        T* Omega_dat = util::upsize(m * k, this->Omega);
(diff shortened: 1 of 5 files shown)


RandLAPACK/drivers/rl_revd2.hh
```diff
@@ -0,0 +1,240 @@
...
+        ///
+        /// @param[in] m
+        ///     The number of rows in the matrix A.
+        ///
+        /// @param[in] n
+        ///     The number of columns in the matrix A.
+        ///
+        /// @param[in] A
+        ///     The m-by-m matrix A, stored in a column-major format.
+        ///     Must be SPD.
```
rileyjmurray, line 68: Should be specified by its upper or lower triangle.

RandLAPACK/drivers/rl_revd2.hh
```diff
@@ -0,0 +1,240 @@
...
+        ///     On entry, is empty and may not have any space allocated for it.
+        ///
+        /// @param[in] E
+        ///     On entry, is empty and may not have any space allocated for it.
+        ///
+        /// @param[out] V
+        ///     Stores m-by-k matrix matrix of eigenvectors.
+        ///
+        /// @param[out] E
+        ///     Stores k eigenvalues.
```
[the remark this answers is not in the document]
rileyjmurray: You'll need to change here and other places, I imagine.

RandLAPACK/drivers/rl_revd2.hh
```diff
@@ -0,0 +1,240 @@
...
+
+        // Construnct a sketching operator
+        // If CholeskyQR is used for stab/orth here, RF fails
+        if(this->RF_Obj.call(m, m, A, k, this->Omega)) {
+            printf("RF/RS FAILED\n");
+            return 2;
+        }
+
+        // Y = A * S
+        blas::gemm(Layout::ColMajor, Op::NoTrans, Op::NoTrans, m, k, m, 1.0, A_dat, m, Omega_dat, m, 0.0, Y_dat, m);
```
rileyjmurray, line 139: ...

Something's wrong here.

Tell me what it is.

RandLAPACK/drivers/rl_revd2.hh
```diff
@@ -0,0 +1,240 @@
...
+        T nu = std::sqrt(m) * std::numeric_limits<double>::epsilon() * lapack::lange(lapack::Norm::Fro, m, k, Y_dat, m);
+
+        // We need Y = Y + vS
+        // We further need R = chol(S' Y)
+        // Solve this as R = chol(S' Y + vS'S)
+        // Compute vS'S - syrk only computes the lower triangular part. Need full.
+        blas::syrk(Layout::ColMajor, Uplo::Lower, Op::Trans, k, m, nu, Omega_dat, m, 0.0, R_dat, k);
+        // Fill the upper triangular part of S'S
+        for(int i = 1; i < k; ++i)
+            blas::copy(k - i, R_dat + i + ((i-1) * k), 1, R_dat + (i - 1) + (i * k), k);
```
rileyjmurray, line 150: Use addressing, not pointer arithmetic.

RandLAPACK/drivers/rl_revd2.hh
```diff
@@ -0,0 +1,240 @@
...
+            // r = number of entries in E that are greater than v
+            if(buf > nu)
+                ++r;
+        }
+
+        // Undo regularlization
+        for(i = 0; i < r; ++i)
+            E[i] -=nu;
+
+        std::fill(V.begin() + m * r, V.end(), 0.0);
```
rileyjmurray, line 186: Remark: I'm okay with using pointer arithmetic here. Not because I like it, but because I dislike the alternatives more.

RandLAPACK/drivers/rl_revd2.hh
rileyjmurray: And then fix any similar mistakes that appear in this function.

RandLAPACK/drivers/rl_revd2.hh
```diff
@@ -0,0 +1,240 @@
...
+        // If CholeskyQR is used for stab/orth here, RF fails
+        if(this->RF_Obj.call(m, m, A, k, this->Omega)) {
+            printf("RF/RS FAILED\n");
+            return 2;
+        }
+
+        // Y = A * S
+        blas::gemm(Layout::ColMajor, Op::NoTrans, Op::NoTrans, m, k, m, 1.0, A_dat, m, Omega_dat, m, 0.0, Y_dat, m);
+
+        T nu = std::sqrt(m) * std::numeric_limits<double>::epsilon() * lapack::lange(lapack::Norm::Fro, m, k, Y_dat, m);
```
rileyjmurray, line 141: Do we really have the ``sqrt(m)`` here when using the Frobenius norm? When I see a ``sqrt(m)`` I think of the worst-case factor by which the Frobenius norm can over-estimate the spectral norm.

RandLAPACK/drivers/rl_revd2.hh
```diff
@@ -0,0 +1,240 @@
...
+            buf = std::pow(S[i], 2);
+            E[i] = buf;
+            // r = number of entries in E that are greater than v
+            if(buf > nu)
+                ++r;
+        }
+
+        // Undo regularlization
+        for(i = 0; i < r; ++i)
+            E[i] -=nu;
```
rileyjmurray, line 184: This might make some entries of ``E`` negative. Make sure that the code which follows accounts for that possibility.

RandLAPACK/drivers/rl_revd2.hh
```diff
@@ -0,0 +1,240 @@
...
+            // Using the forth column of Omega as a buffer for matrix-vector product
+            gemv(Layout::ColMajor, Op::NoTrans, m, m, 1.0, A_dat, m, Omega_dat, 1, 0.0, Omega_dat + 3 * m, 1);
+            // Compute A*g / ||g|| - V*E*V'*g / ||g||
+            // Result is stored in the 4th column of Omega
+            blas::axpy(m, -1.0, Omega_dat + 2 * m, 1, Omega_dat + 3 * m, 1);
+            // Compute (g / ||g||)' * (A*g / ||g|| - V*E*V'*g / ||g||) - this is our measure for the error
+            err = blas::dot(m, Omega_dat, 1, Omega_dat + 3 * m, 1);	
+            // v_0 <- v
+            std::copy(Omega_dat + 3 * m, Omega_dat + 4 * m, Omega_dat);
+        }
```
rileyjmurray, line 226: Have you considered writing a helper function that performs this operation? If you did, then it would be easier to test this code.

test/drivers/test_revd2.cc
```diff
@@ -0,0 +1,111 @@
...
+        // For running QB
+        std::vector<T> A_buf(m * m, 0.0);
+        RandLAPACK::util::gen_mat_type(m, m, A_buf, k, state, mat_type);
+
+        std::vector<T> A(m * m, 0.0);
+
+        // We're using Nystrom, the original must be positive semidefinite
+        blas::syrk(Layout::ColMajor, Uplo::Lower, Op::Trans, m, m, 1.0, A_buf.data(), m, 0.0, A.data(), m);
+        for(int i = 1; i < m; ++i)
+            blas::copy(m - i, A.data() + i + ((i-1) * m), 1, A.data() + (i - 1) + (i * m), m);
```
rileyjmurray, line 36: Plz no pointer arithmetic unless necessary.

test/drivers/test_revd2.cc
```diff
@@ -0,0 +1,111 @@
...
+    // Rank estimation must be 80 - underestimation - starting with very small rank
+    test_REVD2_general<double>(1000, 100, 1, std::make_tuple(0, std::pow(10, 8), false), state, 64, std::pow(10, -13));
+    // Rank estimation must be 80 - underestimation
+    test_REVD2_general<double>(1000, 100, 10, std::make_tuple(0, std::pow(10, 8), false), state, 80, std::pow(10, -13));
+    // Rank estimation must be 60 - overestimation
+    test_REVD2_general<double>(1000, 100, 10, std::make_tuple(0, std::pow(10, 2), false), state, 160, std::pow(10, -13));
+    // Rank estimation must be 160 - slight overestimation
+    test_REVD2_general<double>(1000, 159, 10, std::make_tuple(0, std::pow(10, 2), false), state, 160, std::pow(10, -13));
+    // Numerically rank deficient matrix - expecting rank estimate = m
+    test_REVD2_general<double>(1000, 1000, 10, std::make_tuple(0, std::pow(10, 2), false), state, 1000, std::pow(10, -13));
```
rileyjmurray, line 110: This last test is pretty expensive. Maybe reduce the size to 100-by-100?

Also: don't use a single ``TEST_F`` fixture, use several. If a test fails then we want the gtest log to tell us where we need to look in the code. Right now if one of these tests fail then gtest will be like "well something's broke but good luck my man."

rileyjmurray (changes requested):

RandLAPACK/drivers/rl_revd2.hh
TeachRaccooon: I was actually assuming that A is stored fully.

RandLAPACK/drivers/rl_revd2.hh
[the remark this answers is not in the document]
TeachRaccooon: Fixed.

RandLAPACK/drivers/rl_revd2.hh
TeachRaccooon: The fact that i'm using Omega instead of S?

TeachRaccooon: I was writing comments that signiy what pseudocode lines we're performing.
Var names are different from pseudocode.

RandLAPACK/drivers/rl_revd2.hh
rileyjmurray: I think it's best to not make that assumption. Users with Hermitian / symmetric matrices already expect to work with only one of the triangular parts. It's true that the memory in the opposite triangle is just sitting there, but the user might have reserved that for some kind of workspace.

Having the user specify a triangle also reduces the likelihood of the code being called on malformed input. For example, the call to Cholesky [here](https://github.com/BallisticLA/RandLAPACK/pull/35/files#diff-01caf3ea6e344fc58f135bc5fdaed76173d08a13509a5ef7290bd6741eb4c014R156) might succeed even if ``R`` isn't SPD, so long as the symmetrization of R is SPD. We wouldn't want that to happen.

RandLAPACK/drivers/rl_revd2.hh
TeachRaccooon: Fixed.

RandLAPACK/drivers/rl_revd2.hh
TeachRaccooon: I mean, i could do &V_dat[m * r], &V_dat[m * k].

rileyjmurray: If you already have the ``V_dat`` pointer declared, then I would prefer that.

RandLAPACK/drivers/rl_revd2.hh
TeachRaccooon: You're right, we can drop the sqrt(m) in Fro norm case.

RandLAPACK/drivers/rl_revd2.hh
TeachRaccooon: Fixed.

RandLAPACK/drivers/rl_revd2.hh
TeachRaccooon: Yes, i just don't know where to put it. 
It would be too specialized for utils.
I guess i could make it a part of REVD2 class?

test/drivers/test_revd2.cc
TeachRaccooon: Ok mom.
Fixed.

RandLAPACK/drivers/rl_revd2.hh
rileyjmurray: Hmm. At first, I thought that I wasn't so special. It's just a method for estimating the largest eigenvalue of a PSD matrix. But then I realized that our PSD matrix, ``A - V E V'``, is represented implicitly. That does complicate things.

My suggestion would be to make it a helper function in the ``rl_evd2.hh`` file. No need to attach it to the ``REVD2`` class at this time.

test/drivers/test_revd2.cc
TeachRaccooon: Fixed.

RandLAPACK/drivers/rl_revd2.hh
TeachRaccooon: I do. Fixed.

RandLAPACK/drivers/rl_revd2.hh
TeachRaccooon: Sure, on it.

RandLAPACK/drivers/rl_revd2.hh
TeachRaccooon: Actually, we begin to overestimate rank really badly when we drop sqrt.

RandLAPACK/drivers/rl_revd2.hh
TeachRaccooon: I see. Switched to symm.

RandLAPACK/drivers/rl_revd2.hh
TeachRaccooon: Ran into an issue while doing this & idk why it is happening. May need help.

TeachRaccooon: Fixed.

RandLAPACK/drivers/rl_revd2.hh
TeachRaccooon: So if we drop sqrt(m), nu is 7.389831e-16 for our test cases, which is never achieved by the algorithm, hence overestimation is happening.

TeachRaccooon: Anyway, this modification makes sense. I'll just set tol to 10^-14 here

RandLAPACK/drivers/rl_revd2.hh
```diff
@@ -113,61 +113,59 @@ class REVD2 : public REVD2alg<T> {
...
-
-            char name [] = "A";
-        RandBLAS::util::print_colmaj(m, m, A_dat, name);
         
     for(int i = 0; i < p; ++i) {
-        T g_norm = blas::nrm2(m, Omega_dat, 1);
+        T g_norm = blas::nrm2(m, vector_buf_dat, 1);
         // Compute g = g / ||g|| - we need this because dot product does not take in an alpha
-        std::for_each(Omega_dat, Omega_dat + m, [g_norm](T &entry) { entry /= g_norm;});
+        std::for_each(vector_buf_dat, vector_buf_dat + m, [g_norm](T &entry) { entry /= g_norm;});
```
rileyjmurray, line 139: I think I might be missing something. Can't this be done with ``blas::scal``?

RandLAPACK/drivers/rl_revd2.hh
```diff
@@ -113,61 +113,59 @@ class REVD2 : public REVD2alg<T> {
...
     int64_t m,
     int64_t k,
     int p,
-    T* Omega_dat,
+    T* vector_buf_dat,
     T* V_dat,
     blas::Uplo uplo,
     T* A_dat,
-    T* Y_dat,
+    T* Mat_buf_dat,
```
rileyjmurray: I suggest you remove the "dat" suffixes from the relevant variables here. You have that suffix in the calling function because the calling function obtains these as ``.data()`` of ``std::vector<T>`` variables. Once you factor out a function from its context it's good to rename variables so that the abstract context is here. You already did that by renaming ``Y_dat`` and ``Omega_dat``; let's take it one step further.

test/drivers/test_revd2.cc
```diff
@@ -68,6 +64,7 @@ class TestREVD2 : public ::testing::Test
 
         k = k_start;
         // Regular QB2 call
```
rileyjmurray, line 66: This comment probably shouldn't mention QB2.

rileyjmurray (changes requested):
I only have minor comments. I'd like clarification on one more thing besides the comments in this review. What happened with the ``sqrt(m)`` business in ``nu``? You said that once you removed it you got bad rank over-estimation. But it looks like ``sqrt(m)`` is still removed.

RandLAPACK/drivers/rl_revd2.hh
TeachRaccooon: Fixed.

RandLAPACK/drivers/rl_revd2.hh
TeachRaccooon: Yes, it can be. I'm just no using scal function very often, so didnt think about that.

TeachRaccooon:
> I only have minor comments. I'd like clarification on one more thing besides the comments in this review. What happened with the `sqrt(m)` business in `nu`? You said that once you removed it you got bad rank over-estimation. But it looks like `sqrt(m)` is still removed.

I explained it in the thread where you brought up this issue. Basically, both tol and nu were set to about 10^-16 in case when we drop sqrt(m) from nu. REVD2 never achieves that good of an approximation, so it would estimate rank = m. I dropped sqrt and set tol to 10^-14, and all tests work as expected.

PR #39, opened 2023-05-24
Revise REVD2 computational routines: the rangefinder code path should never use GEMM on a symmetric matrix

rileyjmurray:
This PR introduces two new interfaces and classes: SymmetricPowerSketch (implemented by SYPS) and SymmetricRangefinder (implemented by SYRF). These classes are used in the modified REVD2 implementation. Tests are included which verify that REVD2 doesn't reference invalid memory in the symmetric matrix and that REVD produces the same results when the matrix is lower-triangular vs upper-triangular.

Minor point: we re-learned a painful lesson about the meaning of arguments in LAPACK's ``laswap`` function; the first argument is the number of columns in the matrix, not the number of rows.


TeachRaccooon:
@rileyjmurray everything works now.
One minor thing to note: I have test cases that are supposed to overestimate rank.
You added some assertion check for the estimated rank to be less than the rank we generate the matrix with. I got rid of it.

RandLAPACK/drivers/rl_revd2.hh
```diff
@@ -12,106 +13,99 @@
...
-            verbosity = verb;
+            RandLAPACK::SymmetricRangeFinder<T, RNG>& syrf_obj,
+            int error_est_power_iters,
+            bool verb = false
+        ) : SYRF_Obj(syrf_obj) {
+            error_est_p = error_est_power_iters;
+            verbose = verb;
         }
 
         /// Computes a rank-k approximation to an EVD of a symmetric positive definite matrix:
```
rileyjmurray, line 48: positive semidefinite

test/drivers/test_revd2.cc
```diff
@@ -80,6 +112,25 @@ class TestREVD2 : public ::testing::Test
         norm_A = lapack::lange(Norm::Fro, m, m, all_data.A_cpy.data(), m);
     }
 
+    template <typename T, typename RNG>
+    static void uplo_computational_helper(REVD2UploTestData<T>& all_data) {
+        auto m = all_data.dim;
+        // Filling the upper-triangular matrix
+        blas::syrk(Layout::ColMajor, Uplo::Upper, Op::Trans, m, m, 1.0, all_data.work.data(), m, 0.0, all_data.A_u.data(), m);
+        // Filling the lower-triangular matrix
+        blas::syrk(Layout::ColMajor, Uplo::Lower, Op::Trans, m, m, 1.0, all_data.work.data(), m, 0.0, all_data.A_l.data(), m);
```
rileyjmurray, line 121: @TeachRaccooon, don't use a second call to syrk. Just copy to the other triangle.

rileyjmurray: That is, copy the upper triangle of ``A_u`` into the lower triangle of ``A_l``, instead of writing to the lower triangle of ``A_l`` with ``syrk``.

rileyjmurray:
I have one small comment. Please fix, then go ahead and merge.

rileyjmurray:
The failing macOS build is unrelated. It's an OpenMP configuration problem that I encountered on my Mac recently, and apparently now affects the macOS images on GitHub Actions.

test/drivers/test_revd2.cc
TeachRaccooon: Fixed.

PR #43, opened 2023-05-25
Test expand

TeachRaccooon:
Adding basic tests for RangeFinder and refactoring the existing tests layout to perform space allocation, auxillary computations and the actual test run separately. The changes did not touch [test_determiter.cc](https://github.com/BallisticLA/RandLAPACK/blob/main/test/comps/test_determiter.cc), [test_pcgls.cc](https://github.com/BallisticLA/RandLAPACK/blob/main/test/comps/test_pcgls.cc) and [test_preconditioners.cc](https://github.com/BallisticLA/RandLAPACK/blob/main/test/comps/test_preconditioners.cc).

### DIFF
--- a/test/comps/test_orth.cc
+++ b/test/comps/test_orth.cc
@@ -26,6 +26,9 @@ class TestOrth : public ::testing::Test
 
     template <typename T>
     struct OrthTestData {
+        int64_t row;
+        int64_t col;
+        int64_t rank;
         std::vector<T> A;
         std::vector<T> Y;
         std::vector<T> Omega;
@@ -36,11 +39,20 @@ class TestOrth : public ::testing::Test
             Y(m * k, 0.0),
             Omega(n * k, 0.0),
             I_ref(k * k, 0.0)
-            {}
+            {
+                row = m;
+                col = n;
+                rank = k;
+            }
     };
 
     template <typename T, typename RNG>
-    static void sketch_and_copy_computational_helper(int64_t m, int64_t n, int64_t k, RandBLAS::base::RNGState<RNG> state, OrthTestData<T>& all_data) {
+    static void sketch_and_copy_computational_helper(RandBLAS::base::RNGState<RNG> state, OrthTestData<T>& all_data) {
+
+        auto m = all_data.row;
+        auto n = all_data.col;
+        auto k = all_data.rank;
+
         // Fill the gaussian random matrix
         RandBLAS::dense::DenseDist D{.n_rows = n, .n_cols = k};
         state = RandBLAS::dense::fill_buff(all_data.Omega.data(), D, state);
@@ -56,10 +68,11 @@ class TestOrth : public ::testing::Test
     /// Checks I - \transpose{Q}Q.
     template <typename T, typename RNG>
     static void test_orth_sketch(
-        int64_t m, 
-        int64_t k, 
         OrthTestData<T>& all_data, 
         RandLAPACK::CholQRQ<T>& CholQRQ) {
+
+        auto m = all_data.row;
+        auto k = all_data.rank;
 
         T* Y_dat = all_data.Y.data();
         T* I_ref_dat = all_data.I_ref.data();
@@ -93,6 +106,6 @@ TEST_F(TestOrth, Test_CholQRQ)
     RandLAPACK::CholQRQ<double> CholQRQ(false, false);
 
     RandLAPACK::util::gen_mat_type<double, r123::Philox4x32>(m, n, all_data.A, k, state, std::make_tuple(0, 2, false));
-    sketch_and_copy_computational_helper<double, r123::Philox4x32>(m, n, k, state, all_data);
-    test_orth_sketch<double, r123::Philox4x32>(m, k, all_data, CholQRQ);
+    sketch_and_copy_computational_helper<double, r123::Philox4x32>(state, all_data);
+    test_orth_sketch<double, r123::Philox4x32>(all_data, CholQRQ);
 }
--- a/test/comps/test_orth.cc
+++ b/test/comps/test_orth.cc
@@ -40,7 +40,7 @@ class TestOrth : public ::testing::Test
     };
 
     template <typename T, typename RNG>
-    static void computational_helper(int64_t m, int64_t n, int64_t k, RandBLAS::base::RNGState<RNG> state, OrthTestData<T>& all_data) {
+    static void sketch_and_copy_computational_helper(int64_t m, int64_t n, int64_t k, RandBLAS::base::RNGState<RNG> state, OrthTestData<T>& all_data) {
         // Fill the gaussian random matrix
         RandBLAS::dense::DenseDist D{.n_rows = n, .n_cols = k};
         state = RandBLAS::dense::fill_buff(all_data.Omega.data(), D, state);
@@ -89,6 +89,6 @@ TEST_F(TestOrth, Test_CholQRQ)
     OrthTestData<double> all_data(m, n, k);
 
     RandLAPACK::util::gen_mat_type<double, r123::Philox4x32>(m, n, all_data.A, k, state, std::make_tuple(0, 2, false));
-    computational_helper<double, r123::Philox4x32>(m, n, k, state, all_data);
+    sketch_and_copy_computational_helper<double, r123::Philox4x32>(m, n, k, state, all_data);
     test_orth_sketch<double, r123::Philox4x32>(m, k, all_data);
 }
--- a/test/comps/test_orth.cc
+++ b/test/comps/test_orth.cc
@@ -32,11 +32,11 @@ class TestOrth : public ::testing::Test
         std::vector<T> I_ref;
 
         OrthTestData(int64_t m, int64_t n, int64_t k) :
-        A(m * n, 0.0),
-        Y(m * k, 0.0),
-        Omega(n * k, 0.0),
-        I_ref(k * k, 0.0)
-        {}
+            A(m * n, 0.0),
+            Y(m * k, 0.0),
+            Omega(n * k, 0.0),
+            I_ref(k * k, 0.0)
+            {}
     };
 
     template <typename T, typename RNG>
--- a/test/comps/test_orth.cc
+++ b/test/comps/test_orth.cc
@@ -55,7 +55,7 @@ class TestOrth : public ::testing::Test
     /// Tests orthogonality of a matrix Q, obtained by orthogonalizing a Gaussian sketch.
     /// Checks I - \transpose{Q}Q.
     template <typename T, typename RNG>
-    static void test_orth_sketch(int64_t m, int64_t n, int64_t k, RandBLAS::base::RNGState<RNG> state, OrthTestData<T>& all_data) {
+    static void test_orth_sketch(int64_t m, int64_t k, OrthTestData<T>& all_data) {
 
         T* Y_dat = all_data.Y.data();
         T* I_ref_dat = all_data.I_ref.data();
@@ -90,5 +90,5 @@ TEST_F(TestOrth, Test_CholQRQ)
 
     RandLAPACK::util::gen_mat_type<double, r123::Philox4x32>(m, n, all_data.A, k, state, std::make_tuple(0, 2, false));
     computational_helper<double, r123::Philox4x32>(m, n, k, state, all_data);
-    test_orth_sketch<double, r123::Philox4x32>(m, n, k, state, all_data);
+    test_orth_sketch<double, r123::Philox4x32>(m, k, all_data);
 }
--- a/test/comps/test_orth.cc
+++ b/test/comps/test_orth.cc
@@ -24,41 +24,52 @@ class TestOrth : public ::testing::Test
 
     virtual void TearDown() {};
 
+    template <typename T>
+    struct OrthTestData {
+        std::vector<T> A;
+        std::vector<T> Y;
+        std::vector<T> Omega;
+        std::vector<T> I_ref;
+
+        OrthTestData(int64_t m, int64_t n, int64_t k) :
+        A(m * n, 0.0),
+        Y(m * k, 0.0),
+        Omega(n * k, 0.0),
+        I_ref(k * k, 0.0)
+        {}
+    };
+
+    template <typename T, typename RNG>
+    static void computational_helper(int64_t m, int64_t n, int64_t k, RandBLAS::base::RNGState<RNG> state, OrthTestData<T>& all_data) {
+        // Fill the gaussian random matrix
+        RandBLAS::dense::DenseDist D{.n_rows = n, .n_cols = k};
+        state = RandBLAS::dense::fill_buff(all_data.Omega.data(), D, state);
+        
+        // Generate a reference identity
+        RandLAPACK::util::eye(k, k, all_data.I_ref);
+
+        // Y = A * Omega
+        blas::gemm(Layout::ColMajor, Op::NoTrans, Op::NoTrans, m, k, n, 1.0, all_data.A.data(), m, all_data.Omega.data(), n, 0.0, all_data.Y.data(), m);
+    }
+
     /// Tests orthogonality of a matrix Q, obtained by orthogonalizing a Gaussian sketch.
     /// Checks I - \transpose{Q}Q.
     template <typename T, typename RNG>
-    static void test_orth_sketch(int64_t m, int64_t n, int64_t k, std::tuple<int, T, bool> mat_type, RandBLAS::base::RNGState<RNG> state) {
+    static void test_orth_sketch(int64_t m, int64_t n, int64_t k, RandBLAS::base::RNGState<RNG> state, OrthTestData<T>& all_data) {
 
-        int64_t size = m * n;
-        std::vector<T> A(size, 0.0);
-        std::vector<T> Y(m * k, 0.0);
-        std::vector<T> Omega(n * k, 0.0);
-        std::vector<T> I_ref(k * k, 0.0);
+        T* Y_dat = all_data.Y.data();
+        T* I_ref_dat = all_data.I_ref.data();
 
-        T* A_dat = A.data();
-        T* Y_dat = Y.data();
-        T* Omega_dat = Omega.data();
-        T* I_ref_dat = I_ref.data();
-
-        RandLAPACK::util::gen_mat_type(m, n, A, k, state, mat_type);
-
-        // Fill the gaussian random matrix
-        RandBLAS::dense::DenseDist D{.n_rows = n, .n_cols = k};
-        state = RandBLAS::dense::fill_buff(Omega_dat, D, state);
-        // Generate a reference identity
-        RandLAPACK::util::eye(k, k, I_ref);
-        // Y = A * Omega
-        blas::gemm(Layout::ColMajor, Op::NoTrans, Op::NoTrans, m, k, n, 1.0, A_dat, m, Omega_dat, n, 0.0, Y_dat, m);
         // Orthogonalization Constructor
         RandLAPACK::CholQRQ<T> CholQRQ(false, false);
 
         // Orthonormalize sketch Y
-        if(CholQRQ.call(m, k, Y) != 0) {
+        if(CholQRQ.call(m, k, all_data.Y) != 0) {
             EXPECT_TRUE(false) << "\nPOTRF FAILED DURE TO ILL-CONDITIONED DATA\n";
             return;
         }
         // Call the scheme twice for better orthogonality
-        CholQRQ.call(m, k, Y);
+        CholQRQ.call(m, k, all_data.Y);
         // Q' * Q  - I = 0
         blas::syrk(Layout::ColMajor, Uplo::Upper, Op::Trans, k, m, 1.0, Y_dat, m, -1.0, I_ref_dat, k);
         T norm_fro = lapack::lansy(lapack::Norm::Fro, Uplo::Upper, k, I_ref_dat, k);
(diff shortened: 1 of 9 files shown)


test/comps/test_orth.cc
```diff
@@ -24,41 +24,52 @@ class TestOrth : public ::testing::Test
...
+        state = RandBLAS::dense::fill_buff(all_data.Omega.data(), D, state);
+        
         // Generate a reference identity
-        RandLAPACK::util::eye(k, k, I_ref);
+        RandLAPACK::util::eye(k, k, all_data.I_ref);
+
         // Y = A * Omega
-        blas::gemm(Layout::ColMajor, Op::NoTrans, Op::NoTrans, m, k, n, 1.0, A_dat, m, Omega_dat, n, 0.0, Y_dat, m);
+        blas::gemm(Layout::ColMajor, Op::NoTrans, Op::NoTrans, m, k, n, 1.0, all_data.A.data(), m, all_data.Omega.data(), n, 0.0, all_data.Y.data(), m);
+    }
```
rileyjmurray, line 53: Give this function a more informative name.

test/comps/test_qb.cc
```diff
@@ -15,62 +15,78 @@ class TestQB : public ::testing::Test
...
+        // Get low-rank SVD
+        lapack::gesdd(Job::SomeVec, m, n, all_data.A_cpy.data(), m, all_data.s.data(), all_data.U.data(), m, all_data.VT.data(), n);
+    }
+
+    template <typename T, typename RNG>
+    static void computational_helper(int64_t m, int64_t n, T& norm_A, QBTestData<T>& all_data) {
+
+        // pre-compute norm
+        norm_A = lapack::lange(Norm::Fro, m, n, all_data.A.data(), m);
+    }
```
rileyjmurray, line 68: 1. Don't use function names like this.
2. This function signature is awkward. Why not just return ``norm_A``?
3. This function is trivial. Just set ``all_data.norm_A = lapack::lange(...)`` elsewhere in the code, once you've populated ``all_data.A``.

test/comps/test_qb.cc
```diff
@@ -143,14 +133,12 @@ class TestQB : public ::testing::Test
...
-        // Get low-rank SVD
-        lapack::gesdd(Job::SomeVec, m, n, A_cpy_dat, m, s_dat, U_dat, m, VT_dat, n);
+        // buffer zero vector
         // buffer zero vector
         std::vector<T> z_buf(n, 0.0);
-        T* z_buf_dat = z_buf.data();
         // zero out the trailing singular values
-        blas::copy(n - k, z_buf_dat, 1, s_dat + k, 1);
-        RandLAPACK::util::diag(n, n, s, n, S);
+        blas::copy(n - k, z_buf.data(), 1, s_dat + k, 1);
```
rileyjmurray, line 140: Defining ``z_buf`` and using it in this way is silly. Just use ``blas::scal(n-k, 0.0, s_dat, 1)`` or 
```
for (int i = 0; i < n - k; ++i)
    s_dat[i] = 0.0;
```

test/comps/test_rf.cc
```diff
@@ -0,0 +1,163 @@
...
+    /// General test for QB:
+    /// Computes QB factorzation, and checks:
+    /// 1. A - QB
+    /// 2. B - \transpose{Q}A
+    /// 3. I - \transpose{Q}Q
+    /// 4. A_k - QB = U_k\Sigma_k\transpose{V_k} - QB
+    template <typename T, typename RNG>
+    static void test_RF_general(int64_t m, int64_t n, int64_t k, int64_t p, RandBLAS::base::RNGState<RNG> state, RFTestData<T>& all_data) {
+
+        printf("|==================================TEST QB2 GENERAL BEGIN==================================|\n");
```
rileyjmurray, line 59: This printf says ``QB2`` when it should say ``RF``.

test/comps/test_rf.cc
```diff
@@ -0,0 +1,163 @@
...
+
+        // Make subroutine objects
+        // Stabilization Constructor - Choose PLU
+        RandLAPACK::PLUL<T> Stab(cond_check, verbosity);
+        // RowSketcher constructor - Choose default (rs1)
+        RandLAPACK::RS<T, RNG> RS(Stab, state, p, passes_per_iteration, verbosity, cond_check);
+        // Orthogonalization Constructor - Choose CholQR
+        RandLAPACK::HQRQ<T> Orth_RF(cond_check, verbosity);
+        // RangeFinder constructor - Choose default (rf1)
+        RandLAPACK::RF<T> RF(RS, Orth_RF, verbosity, cond_check);
```
rileyjmurray, line 74: The RangeFinder algorithm object should be defined in the ``TEST_F`` code blocks, not in this function. This function should accept a ``RangeFinder`` as an argument (passed by address).

test/comps/test_util.cc
```diff
@@ -26,68 +26,126 @@ class TestUtil : public ::testing::Test
...
+        std::vector<T> A_norm;
+
+        UtilTestData(int64_t m, int64_t n) :
+        A(m * n, 0.0), 
+        A_cpy(m * n, 0.0), 
+        s(n, 0.0),
+        A_norm(m * n, 0.0) 
+        {}
+    };
+
```
rileyjmurray, line 43: This is kind of weird. I think it's unlikely that all tests of utilities will need ``A_cpy`` or ``A_norm``, yet you're allocating both of them for every test. I see how you're mimicking the structure of ``QBTestData``, but that isn't really appropriate here. Think about the differences in context: QB exists in the context of RandLAPACK's highly configurable algorithms-as-objects API, while utility functions are mundane and handle very different tasks.

test/drivers/test_cqrrpt.cc
```diff
@@ -17,55 +17,83 @@ class TestCQRRPT : public ::testing::Test
...
+        A_cpy(m * n, 0.0) 
+        {}
+    };
 
-        std::vector<T> A_hat(size, 0.0);
-        std::copy(A.data(), A.data() + size, A_hat.data());
+    template <typename T, typename RNG>
+    static void computational_helper(int64_t m, int64_t n, CQRRPTTestData<T>& all_data) {
+        lapack::lacpy(MatrixType::General, m, n, all_data.A.data(), m, all_data.A_cpy.data(), m);
+    }
```
rileyjmurray, line 37: Don't bother with this, my dude. In the place where you call this function, you can just do ``blas::copy(m*n, all_data.A.data(), 1, all_data.A_cpy.data(), 1)``; it's way more readable that way.

test/drivers/test_revd2.cc
rileyjmurray, line 1: This is much better!

rileyjmurray (changes requested):
I didn't look at the driver tests in detail but I gave feedback on other files. Please implement the fixes for those bits of feedback, and apply the "spirit" of those fixes to driver tests as needed.

rileyjmurray (changes requested):
I didn't look at the driver tests in detail but I gave feedback on other files. Please implement the fixes for those bits of feedback, and apply the "spirit" of those fixes to driver tests as needed.

test/comps/test_orth.cc
TeachRaccooon: I kept that name across all file so that the reader can get a feeling for the fact that the testing structure is universal.

test/comps/test_qb.cc
TeachRaccooon: Once again, the only reason i added it is so that different test files have identical structure, but i can change that.

test/comps/test_qb.cc
TeachRaccooon: True, that was just the approach i used when writing this thing last year. Can also just use std::fill.

test/comps/test_rf.cc
TeachRaccooon: My bad. I'm actually getting rid of all those prints.

test/drivers/test_revd2.cc
TeachRaccooon: Sound good!

test/comps/test_orth.cc
TeachRaccooon: Fixed.

test/comps/test_qb.cc
TeachRaccooon: Fixed.

test/comps/test_orth.cc
TeachRaccooon: Fixed.

test/drivers/test_cqrrpt.cc
TeachRaccooon: Fixed.

test/comps/test_util.cc
TeachRaccooon: Fixed.

test/comps/test_qb.cc
```diff
@@ -15,115 +15,108 @@ class TestQB : public ::testing::Test
...
     template <typename T, typename RNG>
-    static void test_QB2_low_exact_rank(int64_t m, int64_t n, int64_t k, int64_t p, int64_t block_sz, T tol, std::tuple<int, T, bool> mat_type, RandBLAS::base::RNGState<RNG> state) {
+    static void test_QB2_low_exact_rank(
+        int64_t m, 
+        int64_t n, 
+        int64_t k,  
+        int64_t block_sz, 
+        T tol,  
+        QBTestData<T>& all_data,
+        algorithm_objects<T, RNG>& all_algs) {
```
rileyjmurray, line 101: I don't think you need ``algorithm_objects`` as an argument. It looks like it suffices to provide the ``QB`` object.

That said, I do like the idea of having a helper function that defines a QB object from the following parameters.
```
    bool verbosity, 
    bool cond_check, 
    bool orth_check, 
    int64_t p, 
    int64_t passes_per_iteration, 
    RandBLAS::base::RNGState<RNG> state
```

TeachRaccooon: A helper function does not work, same issue as we had yesterday.
Upon exist from such function, objects stop existing.
That is why a made a struct.

rileyjmurray: Nested initialization should work.

```
QB(
    bool verbosity, 
    bool cond_check, 
    bool orth_check, 
    int64_t p, 
    int64_t passes_per_iteration, 
    RandBLAS::base::RNGState<RNG> state
) : RF_obj(
        RS_obj(
            Stab_obj(...),
            ....
        ),
        Orth_obj(...),...
    ),
    Orth_obj(...) {}
```
This assumes that the member associated with each class that needs an ``Alg`` object is called ``Alg_obj``, all the way from Orth to Stab.

This actually would amount to a custom constructor for the QB class.

TeachRaccooon: There is a design question related to this. For a given algorithm class, all subroutine algorithm objects have reference types. That is, if something were to change in the subroutine, this will be reflected in the main algorithm. I don't think we're using this feature anywhere at the moment, but it makes sense to be able to use it.
 Because of that, we cannot exactly do a nested decloration - neither GCC allows for it, nor it technically makes sense.

TeachRaccooon: So the only alternatives i see right now are the way i did things in the most recent commit, or just declaring all objects in TEST_F environment and passing the QB object into the test function.

rileyjmurray: Ugh. This is frustrating. Okay, we'll go with your struct solution for now. I'll want to revisit this later.

test/drivers/test_revd2.cc
```diff
@@ -14,69 +14,94 @@ class TestREVD2 : public ::testing::Test
...
 
-    /// General test for RSVD:
-    /// Computes the decomposition factors, then checks A-U\Sigma\transpose{V}.
-    template <typename T, typename RNG>
-    static void test_REVD2_general(int64_t m, int64_t k, int64_t k_start, std::tuple<int, T, bool> mat_type, RandBLAS::base::RNGState<RNG> state, int rank_expectation, T err_expectation) {
+    template <typename T>
+    struct REVD2TestData {
+        std::vector<T> A_buf;
+        std::vector<T> A;
+        std::vector<T> A_cpy;
```
rileyjmurray, line 21: What's the difference between ``A_buf`` and ``A_cpy``?

test/drivers/test_revd2.cc
```diff
@@ -14,69 +14,94 @@ class TestREVD2 : public ::testing::Test
...
-        RandLAPACK::PLUL<T> Stab(cond_check, verbosity);
-        // RowSketcher constructor - Choose default (rs1)
-        RandLAPACK::RS<T, RNG> RS(Stab, state, p, passes_per_iteration, verbosity, cond_check);
-        // Orthogonalization Constructor - Choose HouseholderQR
-        RandLAPACK::HQRQ<T> Orth_RF(cond_check, verbosity);
-        // RangeFinder constructor - Choose default (rf1)
-        RandLAPACK::RF<T> RF(RS, Orth_RF, verbosity, cond_check);
-        // REVD2 constructor
-        RandLAPACK::REVD2<T, RNG> REVD2(RF, std::numeric_limits<double>::epsilon(), 10, state, verbosity);
+                    // Create a copy of the original matrix
```
rileyjmurray, line 70: This comment is indented weirdly far.

test/drivers/test_revd2.cc
```diff
@@ -14,69 +14,94 @@ class TestREVD2 : public ::testing::Test
...
+    struct REVD2TestData {
+        std::vector<T> A_buf;
+        std::vector<T> A;
+        std::vector<T> A_cpy;
+        std::vector<T> V;
+        std::vector<T> eigvals;
+        std::vector<T> E;
+        std::vector<T> Buf;
+
+        REVD2TestData(int64_t m, int64_t k) :
```
rileyjmurray, line 27: It will help you later if you keep track of ``m`` and ``k`` in this struct.

test/drivers/test_revd2.cc
TeachRaccooon: Fair.

TeachRaccooon: Fixed.

test/drivers/test_revd2.cc
TeachRaccooon: Not sure if i understand what you mean by this.

test/drivers/test_revd2.cc
rileyjmurray: If ``A_buf`` is temporary workspace that's only used in one place then I don't think you need to allocate it in advance. Just allocate, use it to define ``A``, then deallocate. (Of course since this is std::vector you don't need explicit deallocation but you get what I mean.)

TeachRaccooon: I get that, will fix.
But this technically gets into the realm of reusing space vs readability.

test/drivers/test_revd2.cc
```diff
@@ -14,69 +14,94 @@ class TestREVD2 : public ::testing::Test
...
+                            bool cond_check, 
+                            int64_t p, 
+                            int64_t passes_per_iteration, 
+                            RandBLAS::base::RNGState<RNG> state) :
+                            Stab(cond_check, verbosity),
+                            RS(Stab, state, p, passes_per_iteration, verbosity, cond_check),
+                            Orth_RF(cond_check, verbosity),
+                            RF(RS, Orth_RF, verbosity, cond_check),
+                            REVD2(RF, state, verbosity)
+                            {}
```
rileyjmurray, line 56: I'll ask that you format this differently for improved readability
```
algorithm_objects(
    bool verbosity, 
    bool cond_check, 
    int64_t p, 
    int64_t passes_per_iteration, 
    RandBLAS::base::RNGState<RNG> state
) : 
    Stab(cond_check, verbosity),
    RS(Stab, state, p, passes_per_iteration, verbosity, cond_check),
    Orth_RF(cond_check, verbosity),
    RF(RS, Orth_RF, verbosity, cond_check),
    REVD2(RF, state, verbosity)
    {}
```

test/drivers/test_revd2.cc
rileyjmurray: Okay nvm I see why that doesn't work well.

Please just use a different variable name here.

test/drivers/test_rsvd.cc
```diff
@@ -17,91 +17,118 @@ class TestRSVD : public ::testing::Test
...
-        
-        printf("|==================================TEST QB2 GENERAL BEGIN==================================|\n");
-
-        // For running QB
-        std::vector<T> A(m * n, 0.0);
-        RandLAPACK::util::gen_mat_type(m, n, A, k, state, mat_type);
-
-        int64_t size = m * n;
+    template <typename T>
+    struct RSVDTestData {
```
rileyjmurray, line 21: Have this struct track parameters ``(m, n, k)``.

test/drivers/test_rsvd.cc
```diff
@@ -17,91 +17,118 @@ class TestRSVD : public ::testing::Test
...
 
+        std::vector<T> A;
         // For results comparison
-        std::vector<T> A_hat(size, 0.0);
-        std::vector<T> A_hat_buf (m * k, 0.0);
-        std::vector<T> A_k(size, 0.0);
-        std::vector<T> A_cpy (m * n, 0.0);
+        std::vector<T> A_hat;
+        std::vector<T> A_hat_buf;
```
rileyjmurray, line 26: Please use a different variable name here.

test/drivers/test_revd2.cc
TeachRaccooon: Fixed.

test/drivers/test_rsvd.cc
```diff
@@ -17,91 +17,118 @@ class TestRSVD : public ::testing::Test
...
+        std::vector<T> s;
+        std::vector<T> S;
+        std::vector<T> U;
+        std::vector<T> VT;
 
-        T* A_hat_dat = A_hat.data();
-        T* A_hat_buf_dat = A_hat_buf.data();
-        T* A_k_dat = A_k.data();
-        T* A_cpy_dat = A_cpy.data();
+        RSVDTestData(int64_t m, int64_t n, int64_t k) :
```
rileyjmurray, line 42: Once RSVDTestData has ``(m, n, k)`` as attributes you'll need to set them here.

test/drivers/test_rsvd.cc
```diff
@@ -17,91 +17,118 @@ class TestRSVD : public ::testing::Test
...
+                            int64_t block_sz,
+                            RandBLAS::base::RNGState<RNG> state) :
+                            Stab(cond_check, verbosity),
+                            RS(Stab, state, p, passes_per_iteration, verbosity, cond_check),
+                            Orth_RF(cond_check, verbosity),
+                            RF(RS, Orth_RF, verbosity, cond_check),
+                            Orth_QB(cond_check, verbosity),
+                            QB(RF, Orth_QB, verbosity, orth_check),
+                            RSVD(QB, verbosity, block_sz)
+                            {}
```
rileyjmurray, line 88: Please format this in the way that I suggested for REVD2.

TeachRaccooon: Fixed.

test/drivers/test_rsvd.cc
TeachRaccooon: Once again, d don't see the point exactly. 
This would also mess with readability, since we use those parameters a lot inside the functions.

test/drivers/test_revd2.cc
rileyjmurray: If ``testdata`` is an ``REVD2TestData`` struct, then it should have fields ``testdata.m, testdata.k``. That way you don't need to pass around those integer parameters separately from ``REVD2TestData`` object.

test/drivers/test_revd2.cc
rileyjmurray: > I get that, will fix.
But this technically gets into the realm of reusing space vs readability.

If you need to reuse space then it's better to use a more open-ended variable name like ``work`` or ``work1``. That way the person reading the code knows "Ah, this might be used for a few different purposes. Better be careful when messing with it."

test/drivers/test_rsvd.cc
TeachRaccooon: Switched to a more descriptive one.

test/drivers/test_revd2.cc
TeachRaccooon: Yes, but every time we use them now we'll have to write testdata.m, which is long and messes with readability. Plus techically those are not data, just variables.

TeachRaccooon: But if you think that this is the right way, i'll do it.

test/drivers/test_rsvd.cc
rileyjmurray: You can unpack the dimension parameters once you enter a function. A-la
```
auto m = testdata.m;
auto n = testdata.n;
auto k = testdata.k;
```

test/drivers/test_revd2.cc
rileyjmurray: Restating from another thread:

> You can unpack dimension parameters on entry to a function, like
```
auto m = testdata.m;
auto n = testdata.n;
auto k = testdata.k;
```

TeachRaccooon: So then all this extra code would serve the sole purpose of passing one object instead of object and variables?

test/drivers/test_rsvd.cc
TeachRaccooon: Added.

test/drivers/test_rsvd.cc
TeachRaccooon: Added.

test/drivers/test_revd2.cc
TeachRaccooon: Resolved.

test/drivers/test_revd2.cc
TeachRaccooon: I actually switched it to A_cpy. Feel like its ok in this case, but at some point we'll need to decide on the space reusal stuff.

rileyjmurray (approved):

PR #45, opened 2023-05-31
Modifications to SVD-based preconditioning

rileyjmurray:
get SVD-based preconditioning to work fully in row-major or fully in column-major (rather than requiring the right singular vectors to be in column-major). Also, implement a simple in-place transpose for square matrices; we use this function to avoid allocating extra workspace in computing preconditioners in the column-major case.

### DIFF
--- a/RandLAPACK/comps/rl_preconditioners.hh
+++ b/RandLAPACK/comps/rl_preconditioners.hh
@@ -18,6 +18,67 @@
 
 namespace RandLAPACK {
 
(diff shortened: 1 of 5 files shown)


rileyjmurray:
@PTNobel please review when you get a chance

PTNobel:
Hi; I'm back from SIAMOP, do you want a post-merge review again?

rileyjmurray:
You're welcome to review if you'd like, but don't feel obligated.

PR #47, opened 2023-06-02
Adding reinforced rank estimation to CQRRPT.

TeachRaccooon:
Simple rank estimation in CQRRPT may cause issue in one of the two metrics:

- if $k$ is overestimated, we loose orthogonality in the $Q$ factor obtained via Cholesky QR.
- If $k$ is underestimated, the quality of final approximation will be insufficient.

We want to find the maximum k for wihich Q remains orthogonal then.
We know that condition number of $Q$ factor form Cholesky QR depends on the condition number of $A^{pre}$.  
Since $A^{pre}$ is already pivoted, we expect the factors obtained from an unpivoted QR factorization of $A^{pre}$ (in our case, Cholesky QR) to not have significant numerical differences from those obtained by pivoted QR. 
Hence, the diagonal entries in the $R$ factor obtained via cholesky QR should have relation to the singular values of $A^{pre}$.
Therefore, we can estimate rank of $A^{pre}$ by looking at the diagonal of $R$.
Furthermore, if Cholesky factorization done via <potrf()> was to fail, this rank re-adjustment would fix the issue by ignoring any garbage data in $R$ (see PR#46 for details).

The initial rank estimation is conrolled via a hardcoded epsilon parameter equaling to 2 * machine_eps^0.95.
This is done with a hope that the estimated rank will be small enough for the $R[:k, :k]$ (adjusted factor retuned from <qrcp()>) to be invertible, but large enough for the reinforced rank estimation to be able to adjust it.

Reinforced rank estimation is controlled by the user-specified <eps> parameter. 
Specifically, k is computed as:
$k = i - 1$ for $i$ such that:
$\max(\mathrm{diag}(R[:i, :i])) / \min(\mathrm{diag}(R[:i, :i])) >= \sqrt{\texttt{tol}/ \epsilon}$.
The right side of the expression comes from the idea that the orthogonality loss in Q is proportional to $\epsilon* \mathrm{cond}(A)^2$, stated [here](https://dl.acm.org/doi/pdf/10.1109/ScalA.2014.11).

I deliberately added a test into test_cqrrpt with a "bad" matrix, where we do need to perform rank re-adjustment.

### DIFF
--- a/RandLAPACK/drivers/rl_cqrrpt.hh
+++ b/RandLAPACK/drivers/rl_cqrrpt.hh
@@ -297,6 +297,12 @@ int CQRRPT<T, RNG>::call(
         // Clear R
         std::fill(R.begin(), R.end(), 0.0);
     }
+        // orth is 14 without
+        // orth is 13 with
+        k = 295;
+        this->rank = 295;
+
+    printf("FIRST RANK ESTIMATION %ld\n", k);
 
     if(this -> timing) {
         rank_reveal_t_stop = high_resolution_clock::now();
@@ -382,6 +388,8 @@ int CQRRPT<T, RNG>::call(
 
     k = new_rank;
     this->rank = new_rank;
+
+    printf("SECOND RANK ESTIMATION %ld\n", k);
 
     blas::trsm(Layout::ColMajor, Side::Right, Uplo::Upper, Op::NoTrans, Diag::NonUnit, m, k, 1.0, R_sp_dat, k, A_dat, m);
 
--- a/RandLAPACK/drivers/rl_cqrrpt.hh
+++ b/RandLAPACK/drivers/rl_cqrrpt.hh
@@ -367,6 +367,21 @@ int CQRRPT<T, RNG>::call(
         return 1;
     }
 
+    // Re-estimate rank after we have the R-factor form Cholesky QR
+    // The strategy here is the same as in naive rank estimation
+    int64_t new_rank = k;
+    for(i = 0; i < k; ++i) {
+        if(std::abs(R_sp_dat[i * k + i]) / std::abs(R_sp_dat[0]) < this->eps) {
+            new_rank = i;
+            break;
+        }
+    }
+    // Beware of that R_sp_dat is k by k and needs to be downsized by rows
+    RandLAPACK::util::row_resize(k, k, R_sp, new_rank);
+
+    k = new_rank;
+    this->rank = new_rank;
+
     blas::trsm(Layout::ColMajor, Side::Right, Uplo::Upper, Op::NoTrans, Diag::NonUnit, m, k, 1.0, R_sp_dat, k, A_dat, m);
 
     if(this -> timing)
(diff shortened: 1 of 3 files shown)


RandLAPACK/drivers/rl_cqrrpt.hh
```diff
@@ -367,6 +368,36 @@ int CQRRPT<T, RNG>::call(
...
+        
+        if(curr_entry > running_max) running_max = curr_entry;
+        if(curr_entry < running_min) running_max = running_min;
+
+        if(running_max / running_min >= std::sqrt(this->eps / std::numeric_limits<T>::epsilon())) {
+            new_rank = i - 1;
+            break;
+        }
+    }
+    //new_rank = 100;
```
rileyjmurray, line 392: Remove this comment

RandLAPACK/drivers/rl_cqrrpt.hh
```diff
@@ -367,6 +368,36 @@ int CQRRPT<T, RNG>::call(
         return 1;
     }
 
+    // Re-estimate rank after we have the R-factor form Cholesky QR.
+    // The strategy here is the same as in naive rank estimation.
+    // This also automatically takes care of any potentical failures in Cholesky factorization.
+    // Note that the diagonal of R_sp_dat may not be sorted, so we need to keep the running max/min
+    // We expect the loss in the orthogonality of Q to be approximately equal to 10^-15 * cond(R_sp_dat)^2
```
rileyjmurray, line 375: > // We expect the loss in the orthogonality of Q to be approximately equal to 10^-15 * cond(R_sp_dat)^2

should be

> // We expect the loss in the orthogonality of Q to be approximately equal to u * cond(R_sp_dat)^2, where u is the unit roundoff for the numerical type T.

test/drivers/test_cqrrpt.cc
```diff
@@ -19,83 +19,172 @@ class TestCQRRPT : public ::testing::Test
...
         std::vector<int64_t> J;
-        std::vector<T> A_cpy;
+        std::vector<T> A_cpy1;
+        std::vector<T> A_cpy2;
+        std::vector<T> I_ref;
 
-        CQRRPTTestData(int64_t m, int64_t n) :
+        CQRRPTTestData(int64_t m, int64_t n, int64_t k) :
         A(m * n, 0.0), 
         J(n, 0.0),
```
rileyjmurray, line 34: ``0.0`` can't be ``int64_t``; initialize ``J`` with ``J(n, 0)``.

RandLAPACK/drivers/rl_cqrrpt.hh
TeachRaccooon: Done.

RandLAPACK/drivers/rl_cqrrpt.hh
TeachRaccooon: Fixed.

PR #49, opened 2023-06-08
Krylov new

TeachRaccooon:
1. Adding an alternative to RangeFinder based on block Krylov method. Current implementation passes RangeFinder tests, but fails to serve as a substitution for a classical range finder in QB scheme.
2. Making sure that random states are properly updated and passes around in functions. The convention is that every time RandBLAS generates matrices, we need to manually update random states. This process revealed a RandBLAS issue described [here](https://github.com/BallisticLA/RandBLAS/issues/57).

Note that in practice, QB with rapidly-computed Frobenius norm stopping criterion does not detect errors smaller than `sqrt(epsillon<T>)`.
This does not seem to ever be an explicit issue whenever using a conventional rangefinder, but is occasionally exposed through use of Krylov rangefinder in situations like the following:
 Then, consider the expression for a QB error indicator described in Section 3.1 of [this](https://arxiv.org/pdf/1606.09402.pdf) paper:

```math
\|A - QB\|_{F} = \sqrt{\|A\|^{2}_{F} - \|B\|^{2}_{F}} = \sqrt{(\|A\|_{F} - \|B\|_{F})(\|A\|_{F} + \|B\|_{F})}.
``` 

In terms of accuracy, the best possible outcome of this expression is dictated by the case when Frobenius norms of $A$ and $B$ differ by $10^{-15}$ (we can only have 15 digits of accuracy after decimal point in double-precision arithmetic).
However, $\sqrt(10^{-15}) \approx 3.1 \cdot 10^{-8}$, which seems to be Our realistic accuracy bound for double precision computations. 
This limitation is also described in Section 3.3 of the aforementioned paper. 
For that reason, in MATLAB's "svdsketch()" routine that uses QB method, the default approximation tolerance is set to `eps(class(A))^(1/4)`.

### DIFF
--- a/RandLAPACK/comps/rl_qb.hh
+++ b/RandLAPACK/comps/rl_qb.hh
@@ -167,6 +167,7 @@ int QB<T>::call(
     if(norm_A == 0.0) {
         // Zero matrix termination
         k = curr_sz;
+        throw std::runtime_error("Zero matrix termination.");
         return 1;
     }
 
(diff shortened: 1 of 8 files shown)


TeachRaccooon:
Waiting for RandBLAS [issue ](https://github.com/BallisticLA/RandBLAS/issues/57) to be resolved before proceeding futher.
It is actually rather interestig that this issue did not cause failure in REVD/QB testing.
Also, in QB the generator is tenchincally not getting updated.

rileyjmurray:
@TeachRaccooon what's the status of this PR? Should we try to merge #58 first?

rileyjmurray:
Closing following discussions with Max. A separate branch has improvements on the stuff in this PR.

PR #51, opened 2023-06-23
Nystrom preconditioning : procedural API and support for linear operators

rileyjmurray:
The main changes in this PR are
1. a new function, called ``nystrom_pc_data``, in ``RandLAPACK/comps/rl_preconditioners.hh``.
2. a new file, called``rl_linops.hh``, in ``RandLAPACK/misc/``.

#### Details on the main changes

The file ``rl_linops.hh`` defines two new classes.
* ``SymmetricLinearOperator`` is an interface. Classes that implement this interface must specify the dimension of their underlying matrix, and overload the ``()`` operator with semantics similar to ``blas::symm``.
* ``ExplicitSymLinOp`` implements the above interface for linear operators that would be represented by a tuple ``(uplo, m, A_buff, lda)`` when working with BLAS.

All functions in the code path for computing Nystrom approximations have been modified to accept ``SymmetricLinearOperator`` objects. Overloads have been provided to retain the original function signatures by constructing an ``ExplicitSymLinOp`` from the linear operator's defining data and then calling the new implementation.

``nystrom_pc_data`` only allows column-major output. One can switch the input matrix to row-major by just switching the claimed triangle that the algorithm is supposed to reference (e.g., row-major access to the lower triangle is the same as column-major access to the upper triangle). If a row-major output is required, then the user will need to do that conversion themselves after ``nystrom_pc_data`` returns.

#### Secondary changes in this PR

1. Creating the ``RandLAPACK/misc`` folder and moving ``rl_gen.hh`` and ``rl_util.hh`` into that folder. Note that I didn't change any namespaces here. I did have to edit some CMake files so that it knows about the files in this folder.
2. Removing the requirement from several classes that fixed-size data is passed std::vector. Instead, they can be passed as plain buffers. We still use std::vector for values like ``(V, eigvals)`` in ``nystrom_pc_data`` since we don't know what sizes we'll need for these on entry to the function.
3. Cleaning up include statements.
4. Fixing string formatting in various calls to printf, to resolve compiler warnings.


RandLAPACK/comps/rl_preconditioners.hh
```diff
@@ -217,5 +219,101 @@ int64_t make_right_orthogonalizer(
...
+ * @param[in] mu_min
+ *      The smallest value of mu for which we want (V, eigvals) to define a useful
+ *      preconditioner for regularized linear systems (A + mu*I)x = b.
+ * @param[in] state
+ *      The RNGState used for random sketching inside the algorithm.
+ * 
+ * @param[in] num_syps_passes
+ *      A very small nonnegative integer. Optional; defaults to 3. 
+ *      This controls the number of power iterations used in computing a sketch of A.
+ *      Increasing the value would reduce |A - V diag(eigvals) V'|| when the number
```
PTNobel, line 269: missing a `|` after reduce.

RandLAPACK/comps/rl_preconditioners.hh
```diff
@@ -217,5 +219,101 @@ int64_t make_right_orthogonalizer(
...
+    //      (*) Use power sketching followed by Householder orthogonalization.
+    //      (*) Do not check condition numbers or log to std::out.
+    RandLAPACK::REVD2<T, RNG> NystromAlg(SYRF, num_steps_power_iter_error_est, false);
+    // ^ Define the algorithm for low-rank approximation via Nystrom.
+    //      (*) Handle accuracy requests by estimating ||A - V diag(eigvals) V'||
+    //          with "num_steps_power_iter_error_est" steps of power iteration.
+    //      (*) Do not log to std::out.
+    T tol = mu_min / 5;
+    // ^ Set tolerance to something materially smaller than the smallest
+    //   regularization parameter the user claims to need.
```
PTNobel, line 314: I don't have quite enough context to know if this makes sense, but `/ 5` is materially smaller?

RandLAPACK/comps/rl_syps.hh
PTNobel, line 1: Stellar.

RandLAPACK/comps/rl_syrf.hh
```diff
@@ -30,6 +30,16 @@ class SymmetricRangeFinder {
             RandBLAS::RNGState<RNG>& state,
             T* work_buff
         ) = 0;
+
+        virtual RandBLAS::RNGState<RNG> call(
+            blas::Uplo uplo,
+            int64_t m,
+            const T* A,
```
PTNobel, line 37: <3

RandLAPACK/drivers/rl_revd2.hh
PTNobel, line 1: I'm not closely reviewing these changes fyi; lmk if i should

test/comps/test_preconditioners.cc
PTNobel, line 1: I skimmed this; I don't follow what all the loops are doing and I did a lot of lecturing and office hours today so I don't have the focus to work it out. a few more comments or me coming back tomorrow would probably be good.

PTNobel (changes requested):
Overall, looks like a good implementation but seems to only work for gram matrices, which is ... not very useful in my view.

At least one minor typo in the docs.

RandLAPACK/comps/rl_preconditioners.hh
```diff
@@ -217,5 +219,101 @@ int64_t make_right_orthogonalizer(
...
 }
 
+/**
+ * Construct data (V, eigvals) for a Nystrom preconditioner of regularized linear
+ * systems involving a PSD matrix A.
+ * 
+ *      If our regularized linear system is 
+ *          (A + mu*I)x= b,
+ *      then the preconditioner would be
+ *          P^{-1} = V diag(eigvals + mu)^{-1} V' + (I - VV') / (min(eigvals) + mu).
```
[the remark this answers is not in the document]
rileyjmurray: This was just the notation from the relevant section of the RandLAPACK book. I chose the notation there pretty carefully but perhaps I was overthinking things.

RandLAPACK/comps/rl_preconditioners.hh
```diff
@@ -217,5 +219,101 @@ int64_t make_right_orthogonalizer(
...
+ * 
+ * Optional arguments to this function can trade-off greater computational cost
+ * with (1) better approximations A_hat = V diag(eigvals) V' for fixed k and (2)
+ * better estimates of the spectral norm error ||A - A_hat||.
+ * 
+ * @param[in] uplo
+ *      blas::Uplo::Lower or blas::Uplo::Upper. Specifies the triangular
+ *      part of A that we access while computing (V, eigvals).
+ * @param[in] A
+ *      A buffer for an m \times m positive semidefinite matrix A.
```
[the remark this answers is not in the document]
rileyjmurray: What's this about Gram matrices? Every PSD matrix is indeed a Gram matrix of some other matrix, but that doesn't show up here. Do you mean to ask if ``A`` is an explicit array, rather than a positive semidefinite linear operator?

RandLAPACK/comps/rl_preconditioners.hh
rileyjmurray: It is indeed!

RandLAPACK/comps/rl_preconditioners.hh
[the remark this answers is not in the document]
PTNobel: That's fine; I just don't like the inverse personally :(

RandLAPACK/comps/rl_preconditioners.hh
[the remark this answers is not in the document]
PTNobel: Ah sorry yeah I meant that we're requiring `A` be an explicit PSD matrix; but that seems really disappointing as someone making a least squares solver with this who would prefer to never construct a Gram matrix.

rileyjmurray:
Ah, right! I can add support for linear operators tomorrow.

RandLAPACK.hh
rileyjmurray, line 1: No need to review changes to this file.

RandLAPACK/CMakeLists.txt
rileyjmurray, line 1: No need to review changes to this file.

RandLAPACK/comps/rl_orth.hh
rileyjmurray, line 1: No need to review changes to this file.

RandLAPACK/comps/rl_rf.hh
rileyjmurray, line 1: No need to review changes to this file.

RandLAPACK/drivers/rl_hqrrp.hh
rileyjmurray, line 1: No need to review changes to this file.

RandLAPACK/drivers/rl_cqrrpt.hh
rileyjmurray, line 1: No need to review changes to this file.

RandLAPACK/comps/rl_preconditioners.hh
rileyjmurray: fixed

RandLAPACK/comps/rl_preconditioners.hh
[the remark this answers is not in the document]
rileyjmurray: @PTNobel I've fixed this with a very simple linear operator interface. The size of the PR expanded but you can skip over many of the changed files.

RandLAPACK/comps/rl_syps.hh
```diff
@@ -22,22 +23,19 @@ class SymmetricPowerSketch {
...
-            int64_t m,
-            const T* A,
-            int64_t lda,
+        virtual int call(
+            SymmetricLinearOperator<T> &A,
             int64_t k,
-            RandBLAS::RNGState<RNG> state,
-            T* skop_buff = nullptr,
+            RandBLAS::RNGState<RNG>& state,
+            T*& skop_buff = nullptr,
```
PTNobel, line 38: I'm not sure what a reference to a pointer does, would appreciate some explanation (not necessarily in the code; just so I understand what is happening).

RandLAPACK/comps/rl_syps.hh
```diff
@@ -102,22 +100,19 @@ class SYPS : public SymmetricPowerSketch<T, RNG> {
         int call(
             blas::Uplo uplo,
             int64_t m,
-            const std::vector<T>& A,
+            const T* A,
             int64_t lda,
             int64_t k,
-            RandBLAS::RNGState<RNG>& state,
+            RandBLAS::RNGState<RNG> &state,
```
PTNobel, line 106: You are inconsnistent about where the & goes... you have `T*& ident` below, but here we have `<RNG> &state`

RandLAPACK/misc/rl_linops.hh
```diff
@@ -0,0 +1,79 @@
...
+    SymmetricLinearOperator(int64_t m) : m(m) {};
+
+    /* The semantics of this function are similar to blas::symm.
+        * We compute
+        *      C = alpha * A * B + beta * C
+        * where this SymmetricLinearOperator object represents "A"
+        * and "B" has "n" columns.
+    */
+    virtual void operator()(
+        blas::Layout layout,
```
PTNobel, line 31: This defines the layout of A, B, C?

PTNobel (approved):
This looks mostly great to me! Left a few small comments.

RandLAPACK/comps/rl_syps.hh
rileyjmurray: @TeachRaccooon made this change. It's so we can allocate memory by an assignment ``skop_buff = new T[...]`` inside the function if ``skop_buff`` is passed in as a null pointer. We end up attaching the memory to a sketching operator that deletes it later on. (How it knows how much to delete, I don't know ... but our address sanitizer (valgrind equivalent) shows no problems with this approach.

RandLAPACK/comps/rl_syps.hh
rileyjmurray: I'll file an issue about this so we can address it in a separate PR.

RandLAPACK/misc/rl_linops.hh
rileyjmurray: Only of ``B, C``. Since we're working with a linear operator it's artificial to impose a layout specification for ``A``.

rileyjmurray: I've updated the documentation here. I've also updated the implementation of ExplicitSymLinOp's ``operator()`` function so that it can reconcile differences in the layout of its internal buffer and the layout of ``(B, C)`` that might be passed when the ExplicitSymLinOp is called.

rileyjmurray:
The failing GitHub Actions build for linux is due to problems on GitHub's side. I'll wait until they announce a fix before re-running those tests. If those tests pass, then I'll merge.

PR #57, opened 2023-06-29
Style for white space when passing by address

### LINKED ISSUE #54: style: where to put & when passing by address
In one of his reviews of PR #51, @PTNobel pointed out style inconsistency in our codebase.

If we pass a variable ``var`` of type ``T`` by address, we sometimes write ``T& var`` and sometimes write ``T &var``. This makes no difference from a correctness standpoint, but it does us no favors for readability. My preference is ``T &var, `` but I'm open to discussion on this.

In any case, we should have a dedicated PR for resolving this inconsistency. Maybe we can also adopt a C++ linter?

rileyjmurray:
Resolves #54. We now use the format ``T &var``, with a single space between the type and the ampersand and no space between the ampersand and variable name.

### DIFF
--- a/RandLAPACK/comps/rl_determiter.hh
+++ b/RandLAPACK/comps/rl_determiter.hh
@@ -21,7 +21,7 @@ void pcg(
     const T* b, // length m
     const T* c, // length n
     T delta, // >= 0
-    std::vector<T>& resid_vec, // re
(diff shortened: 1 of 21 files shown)


rileyjmurray:
Only the macOS build is running. It passed the basic build and is now doing ASAN, which passes on my machine. Merging now.

PR #59, opened 2023-10-16
Create CITATION.cff

TeachRaccooon:
Added a citation file for convenience.

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -1,0 +1,8 @@
+@software{RandLAPACK2023,
+  author = {Maksim, Melnichenko and Riley, Murray and Burlen, Loring and Parth, Nobel},
+  month = {10},
+  title = {{RandLAPACK Library}},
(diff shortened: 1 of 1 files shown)


TeachRaccooon:
Build fails bc repo needs updates. I included those in PR 58.

CITATION.cff
```diff
@@ -0,0 +1,8 @@
+@software{RandLAPACK2023,
+  author = {Maksim, Melnichenko and Riley, Murray and Burlen, Loring and Parth, Nobel},
```
rileyjmurray, line 2: @TeachRaccooon only put comments between first and last names if the first name appears _after_ the last name. So "Murray, Riley" is correct but "Riley, Murray" is not.

TeachRaccooon: Fixed.

TeachRaccooon:
@rileyjmurray can i merge this?

PR #92, opened 2024-12-11
BQRRP & CQRRPT Polishing

TeachRaccooon:
This PR is responsible for ensuring that BQRRP (Formerly CQRRP/ICQRRP/CQRRP_blocked) is in-line with the upcoming manuscript, additionally cleaning up the benchmarking code. 
It also introduces a slightly altered file format for the benchmark output files.
It also includes some minor naming updates to RBKI benchmarks.


TeachRaccooon:
@rileyjmurray this is our BQRRP paper PR. I still need to update the GPU benchmark, but otherwise, this should be it for the most part.
Note the slightly updated file format for the benchmark outputs; please let me know if it looks ok.
If so, I will modify the MATLAB plotting scripts accordingly.

TeachRaccooon:
@rileyjmurray pinging you here to request comments.

RandLAPACK/drivers/rl_bqrrp.hh
```diff
@@ -34,46 +34,43 @@ class CQRRPalg {
...
-            use_qp3      = false;
-            use_gemqrt   = false;
-            internal_nb  = b_sz;
-            tol = std::numeric_limits<T>::epsilon();
-            use_qp3      = false;
-            use_gemqrt   = false;
-            internal_nb  = b_sz;
-            tol = std::numeric_limits<T>::epsilon();
+            timing          = time_subroutines;
+            eps             = ep;
```
rileyjmurray, line 62: Looks like "eps" is not used.

rileyjmurray (changes requested):
Request: search for the expressions ``//lapack::`` and ``//blas::`` in the code. That will identify places where you might have disabled code that should just be removed.

Request: please rename the PR to indicate that it also cleans up lots of benchmarking code, including benchmarking code for CQRRPT.

Question: do you really need to include RBKI changes in this PR?

Comment: I'm okay with the benchmarks including updates to BLAS benchmarks, but IMO these would be better in their own PR. (In particular, they can be in their own PR which I don't need to review.)

TeachRaccooon:
> Request: search for the expressions `//lapack::` and `//blas::` in the code. That will identify places where you might have disabled code that should just be removed.
> 
> Request: please rename the PR to indicate that it also cleans up lots of benchmarking code, including benchmarking code for CQRRPT.
> 
> Question: do you really need to include RBKI changes in this PR?
> 
> Comment: I'm okay with the benchmarks including updates to BLAS benchmarks, but IMO these would be better in their own PR. (In particular, they can be in their own PR which I don't need to review.)

RandLAPACK/drivers/rl_bqrrp.hh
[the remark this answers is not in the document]
TeachRaccooon: True; it was used back when we had specialized termination criteria.

RandLAPACK/drivers/rl_bqrrp.hh
```diff
@@ -273,13 +262,13 @@ int CQRRP_blocked<T, RNG>::call(
     // Is of size n * d, with an lda n.
     T* A_sk_trans = ( T * ) calloc( n * d, sizeof( T ) );
 
-    // Buffer for the R-factor in Cholesky QR, of size b_sz by b_sz, lda b_sz.
+    // Buffer for the R-factor in tall QR, of size b_sz by b_sz, lda b_sz.
     // Also used to store the proper R11_full-factor after the 
-    // full Q has been restored form economy Q (the has been found via Cholesky QR);
+    // full Q has been restored form economy Q (the has been found via tall QR);
```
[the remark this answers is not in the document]
TeachRaccooon: Fixed.

benchmark/bench_RBKI/RBKI_speed_comparisons.cc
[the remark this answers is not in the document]
TeachRaccooon: Yeee.

RandLAPACK/misc/rl_util.hh
```diff
@@ -459,4 +461,21 @@ void rl_orhr_col(
...
+
+    // Create a string stream to format the date
+    std::ostringstream dateStream;
+    dateStream << std::setw(4) << std::setfill('0') << (1900 + localTime->tm_year) << "_"  // Year
+               << std::setw(2) << std::setfill('0') << (localTime->tm_mon + 1) << "_"      // Month
+               << std::setw(2) << std::setfill('0') << localTime->tm_mday << "_";          // Day
+
+    return dateStream.str();
+}
+
```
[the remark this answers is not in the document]
TeachRaccooon: Sometimes, I try lol.
This is used for more precise benchmark filenames.

RandLAPACK/drivers/rl_bqrrp_gpu.hh
```diff
@@ -334,12 +337,14 @@ int CQRRP_blocked_GPU<T, RNG>::call(
         cudaMemsetAsync(J_buffer,    (T) 0.0, n, strm);
         cudaMemsetAsync(Work2,       (T) 0.0, n, strm);
 
-        // Perform pivoted LU on A_sk', follow it up by unpivoted QR on a permuted A_sk.
-        // Get a transpose of A_sk
+
         if(this -> timing) {
             nvtxRangePushA("qrcp");
-            qrcp_t_start = high_resolution_clock::now();
+            qrcp_wide_t_start = high_resolution_clock::now();
```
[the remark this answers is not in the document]
TeachRaccooon: Fixed.

RandLAPACK/drivers/rl_bqrrp_gpu.hh
```diff
@@ -42,34 +42,38 @@ class CQRRP_GPU_alg {
...
-        CQRRP_blocked_GPU(
+        BQRRP_GPU(
             bool time_subroutines,
             T ep,
             int64_t b_sz
         ) {
-            timing = time_subroutines;
-            eps = ep;
+            timing     = time_subroutines;
+            eps        = ep;
```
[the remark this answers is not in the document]
TeachRaccooon: Fixed.

RandLAPACK/drivers/rl_bqrrp.hh
```diff
@@ -273,13 +262,13 @@ int CQRRP_blocked<T, RNG>::call(
...
     // That is done by applying the sign vector D from orhr_col().
     // Eventually, will be used to store R11 (computed via trmm)
     // which is then copied into its appropriate space in the matrix A.
-    T* R_cholqr = ( T * ) calloc( b_sz_const * b_sz_const, sizeof( T ) );
+    T* R_tall_qr = ( T * ) calloc( b_sz_const * b_sz_const, sizeof( T ) );
     // Pointer to matrix T from orhr_col at currect iteration, will point to Work2 space.
     T* T_dat    = ( T * ) calloc( b_sz_const * b_sz_const, sizeof( T ) );
```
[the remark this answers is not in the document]
TeachRaccooon: Yeah, its too small of a space to give it a special treatment tbh.

RandLAPACK/drivers/rl_bqrrp.hh
```diff
@@ -122,32 +119,30 @@ class CQRRP_blocked : public CQRRPalg<T, RNG> {
...
         // Naive rank estimation parameter;
         T tol;
+
+        // 10 entries - logs time for different portions of the algorithm
+        std::vector<long> times;
+
+        // Core subroutines options, controlled by user
+        std::string qrcp_wide;     // Supported options: "qp3," "luqr"
+        std::string qr_tall;       // Supported options: "geqrt," "cholqr," "geqrf"
+        std::string apply_trans_q; // Supported options: "gemqrt," "ormqr"
```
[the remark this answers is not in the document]
TeachRaccooon: Fixed.

RandLAPACK/drivers/rl_bqrrp_gpu.hh
```diff
@@ -123,20 +127,24 @@ class CQRRP_blocked_GPU : public CQRRP_GPU_alg<T, RNG> {
...
         int64_t block_size;
+
+        // 15 entries - logs time for different portions of the algorithm
         std::vector<long> times;
 
         // Naive rank estimation parameter;
         T tol;
+
+        // Core subroutines options, controlled by user
+        std::string qr_tall;
```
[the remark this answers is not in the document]
TeachRaccooon: Fixed.

TeachRaccooon:
> Request: search for the expressions `//lapack::` and `//blas::` in the code. That will identify places where you might have disabled code that should just be removed.
> 
> Request: please rename the PR to indicate that it also cleans up lots of benchmarking code, including benchmarking code for CQRRPT.
> 
> Question: do you really need to include RBKI changes in this PR?
> 
> Comment: I'm okay with the benchmarks including updates to BLAS benchmarks, but IMO these would be better in their own PR. (In particular, they can be in their own PR which I don't need to review.)

@rileyjmurray fixed all the things you've pointed out.
I added a note in PR description regarding benchmarks cleanup.
There was no strict need to include RBKI stuff here, but I figured I'll update all benchmark names at once.

rileyjmurray:
@TeachRaccooon your current implementation of enums pollutes the RandLAPACK namespace. I suggest you follow the advice in this StackOverflow post https://stackoverflow.com/a/4571878/2664946, under the heading **Introducing another scope inside a class scope**. Of the two approaches mentioned there I'd use the one of the form
```c++
class Product {
public:
    struct ProductType {
        enum Enum { FANCY, AWESOME, USEFUL };
    };
    struct ProductBoxType {
        enum Enum { BOX, BAG, CRATE };
    };
    Product(ProductType::Enum t, ProductBoxType::Enum b, String name);

    // the rest of the class: fields, methods
};
```
where it would be called like 
```c++
Product p(Product::ProductType::FANCY, Product::ProductBoxType::BOX);
// or ...
using Product::ProductType;
using Product::ProductBoxType;
Product q(ProductType::FANCY, ProductBoxType::BOX);
```

rileyjmurray:
GitHub is also telling me that this needs to have changes from main merged in or to rebase.

TeachRaccooon:
> @TeachRaccooon your current implementation of enums pollutes the RandLAPACK namespace. I suggest you follow the advice in this StackOverflow post https://stackoverflow.com/a/4571878/2664946, under the heading **Introducing another scope inside a class scope**. Of the two approaches mentioned there I'd use the one of the form
> 
> ```c++
> class Product {
> public:
>     struct ProductType {
>         enum Enum { FANCY, AWESOME, USEFUL };
>     };
>     struct ProductBoxType {
>         enum Enum { BOX, BAG, CRATE };
>     };
>     Product(ProductType::Enum t, ProductBoxType::Enum b, String name);
> 
>     // the rest of the class: fields, methods
> };
> ```
> 
> where it would be called like
> 
> ```c++
> Product p(Product::ProductType::FANCY, Product::ProductBoxType::BOX);
> // or ...
> using Product::ProductType;
> using Product::ProductBoxType;
> Product q(ProductType::FANCY, ProductBoxType::BOX);
> ```

Fixed. Lmk if the readability lvl is ok with this change.

RandLAPACK/drivers/rl_bqrrp.hh
```diff
@@ -18,6 +18,11 @@ namespace RandLAPACK {
 template <typename T, typename RNG>
 class BQRRPalg {
     public:
+        struct SubroutineType{
+            enum qrcp_wide_subroutines {luqr, geqp3};
+            enum qr_tall_subroutines {geqrt, cholqr, geqrf};
+            enum apply_trans_q_subroutines {ormqr, gemqrt};
+        };
```
rileyjmurray, line 25: I suggest
```c++
struct Options {
    struct QRCPWide    { enum Enum { LUQR,  QP3           }; };
    struct QRTall      { enum Enum { QRT,   CholQR, QRF   }; };
    struct ApplyTransQ { enum Enum { ORMQR, GEMQRT        }; };
};
```
You could change ``Options`` to ``Subroutines`` if you'd like.

RandLAPACK/drivers/rl_bqrrp.hh
```diff
@@ -55,17 +60,15 @@ class BQRRP : public BQRRPalg<T, RNG> {
...
-            eps             = ep;
             tol             = std::numeric_limits<T>::epsilon();
             block_size      = b_sz;
             internal_nb     = b_sz;
-            qrcp_wide       = "luqr";
-            qr_tall         = "geqrf";
-            apply_trans_q   = "ormqr";
+            qrcp_wide       = BQRRP<T, RNG>::SubroutineType::qrcp_wide_subroutines::luqr;
+            qr_tall         = BQRRP<T, RNG>::SubroutineType::qr_tall_subroutines::geqrf;
+            apply_trans_q   = BQRRP<T, RNG>::SubroutineType::apply_trans_q_subroutines::ormqr;
```
rileyjmurray, line 71: You don't need to fully qualify the scopes here. With your current implementation it should suffice to have lines like the following.
```c++
qrcp_wide = SubroutineType::qrcp_wide_subroutines::luqr;
```
With my suggested implementation it should suffice to have lines like this:
```c++
qrcp_wide = Options::QRCPWide::LUQR
```

RandLAPACK/drivers/rl_bqrrp.hh
```diff
@@ -428,7 +430,7 @@ int BQRRP<T, RNG>::call(
                 qr_tall_t_dur  += duration_cast<microseconds>(qr_tall_t_stop - qr_tall_t_start).count();
                 apply_transq_t_start = high_resolution_clock::now();
             }
-        } else if (this -> qr_tall == "cholqr") {
+        } else if (this -> qr_tall == BQRRP<T, RNG>::SubroutineType::qr_tall_subroutines::cholqr) {
```
rileyjmurray, line 433: As before, you don't need the fully qualified name (prefix ``BQRRP<T, RNG>::`` shouldn't be needed).

test/drivers/test_bqrrp.cc
```diff
@@ -154,12 +154,11 @@ TEST_F(TestBQRRP, BQRRP_full_rank_basic) {
...
     int64_t b_sz = 500;//500;
     double norm_A = 0;
-    double tol = std::pow(std::numeric_limits<double>::epsilon(), 0.85);
     auto state = RandBLAS::RNGState();
 
     BQRRPTestData<double> all_data(m, n, k);
-    RandLAPACK::BQRRP<double, r123::Philox4x32> BQRRP(true, tol, b_sz);
-    BQRRP.qr_tall = "cholqr";
+    RandLAPACK::BQRRP<double, r123::Philox4x32> BQRRP(true, b_sz);
+    BQRRP.qr_tall = RandLAPACK::BQRRP<double, r123::Philox4x32>::SubroutineType::qr_tall_subroutines::cholqr;
```
rileyjmurray, line 161: At the top of this file just toss a``using RandLAPACK::BQRRP::Options`` (or ``using RandLAPACK::BQRRP::Subroutines``, depending on how you name things). Then lines like this can be less verbose.

rileyjmurray (changes requested):

rileyjmurray:
> > GitHub is also telling me that this needs to have changes from main merged in or to rebase.
> 
> Not sure if I've dealt with this the right way. Please verify.

Looks like you fixed it.

RandLAPACK/drivers/rl_bqrrp.hh
TeachRaccooon: Thought so as well, but i get compiler errors if i don't use the full scope.

RandLAPACK/drivers/rl_bqrrp.hh
TeachRaccooon: Same comment as before

RandLAPACK/drivers/rl_bqrrp.hh
TeachRaccooon: Any particular reason for that change or is it just your stylistic preference?

rileyjmurray: 1. We use CamelCase for type names.
2. It's shorter.

test/drivers/test_bqrrp.cc
TeachRaccooon: Tried that, result sin errors.

RandLAPACK/drivers/rl_bqrrp.hh
rileyjmurray: Hmm. Well, you can throw a using statement at the top of the class definition, like I do here in RandBLAS:
https://github.com/BallisticLA/RandBLAS/blob/dea0ae09e84ea20a1ffec757cf09e33b32ac1ab3/RandBLAS/dense_skops.hh#L337-L347

RandLAPACK/drivers/rl_bqrrp.hh
TeachRaccooon: AIt, will rework everything then

test/drivers/test_bqrrp.cc
TeachRaccooon: Ah wait, works with a different syntax.

rileyjmurray: kewl

RandLAPACK/drivers/rl_bqrrp.hh
TeachRaccooon: Downside: three "using" statements instead of one in each file.

rileyjmurray: If you're using them multiple times in one file I don't think that's a significant downside. You can also just do
```c++
using BQO = BQRRP::Options;

// ... later ...
auto luqr = BQO::QRCPWide::LUQR; 
auto qrf = BQO::QRTall::QRF;
```

TeachRaccooon: I feel like using three things instead of one twice (three structs instead of one and three "using" statements instead of one) is still suboptimal, but if you say it is best, I'll do it.

RandLAPACK/drivers/rl_bqrrp.hh
TeachRaccooon: Done

RandLAPACK/drivers/rl_bqrrp.hh
TeachRaccooon: Fixed.

RandLAPACK/drivers/rl_bqrrp.hh
rileyjmurray: The advantage of the nested version is that function signatures accepting these arguments become clearer.

For example,
```c++
template <typename T>
void bqrrp_functional(
    int64_t m, int64_t n, T* A, int64_t lda, T* tau, int64_t* piv,
    Options::QRTall::Enum      qr_tall       = Options::QRTall::QRF,
    Options::QRCPWide::Enum    qrcp_wide     = Options::QRCPWide::LUQR, 
    Options::ApplyTransQ::Enum apply_trans_q = Options::ApplyTransQ::ORMQR
) {
    // Step 1. Construct the BQRRP object and an RNGState
    RNGState s(0);
    // Step 2. invoke the BQRRP object's "call" function.
    // Step 3. ... return.
}
```
That said, there are alternatives here. Maybe ...
```c++
struct BQRRPOptions {
    struct QRCPWide {enum Enum {LUQR, QP3}; };
    struct QRTall   {enum Enum {CholQR, QRF}; };
    QRCPWide::Enum qrcp_wide = QRCPWide::LUQR;
    QRTall::Enum qr_tall = QRTall::QRF;
};
// ^ Can define that inside or outside of the class. Note that it's a type which contains structs and can be instantiated.

template <typename T>
void bqrrp_functional(
    int64_t m, int64_t n, T* A, int64_t lda, T* tau, int64_t* piv, BQRRPOptions ops = {}
) {
    // Step 1. Construct the BQRRP object and an RNGState
    RNGState s(0);
    // Step 2. invoke the BQRRP object's "call" function.
    // Step 3. ... return.
}

// assume you have some A, piv, m, n, and tau defined ...
bqrrp_functional(m, n, A, m, tau, piv);
// ^ the compiler injects an argument BQRRPOptions ops = {}, which
//   which has ops.qrcp_wide and ops.qr_tall automatically initialized to their defaults.

using Options = BQRRPOptions;
// ^ Can include that statement in places where it's the meaning of "Options" is unambiguous

bqrrp_functional(m, n, A, m, tau, {Options::QRCPWide::QP3, Options::QRTall::CholQR}); 
// ^ being explicit about options without constructing a BQRRPOptions object on another line.

```

rileyjmurray: @TeachRaccooon you can mess with alternatives here: 

https://godbolt.org/#z:OYLghAFBqd5QCxAYwPYBMCmBRdBLAF1QCcAaPECAMzwBtMA7AQwFtMQByARg9KtQYEAysib0QXAEx8BBAKoBnTAAUAHpwAMvAFYTStJg1DIApJIBCZ86SX1kBPAMqN0AYVS0AriwZ6XAGTwGTAA5bwAjTGIJAE5SAAdUBUJHBncvHz1E5IcBQOCwlkjorjjbTHtUoQImYgJ0718uG0w7XIZq2oJ80Iio2JsauobM5oUh7qDeov7SgEobVE9iZHYOEw0AQQ3N8eJPewBqcwBFACUz5QB5ePaFQ5MAdkstw7fDvYOCQ/PXZQB1PBYB7PRjeQ7YBjgp7mfxyc6kH7KADMTwAIiZkeYQRisTt3h8CPsjucACpiWjvGFglgQqG0mGuBAeBE/M4AMXRmOxXLxr3evwBQPYIEh4IAjit4gB9ADuwoeyLRbL%2BgKwIBAcPO3PxArO5NotA1YtpkulNUNiuVZIpGvOnL520euJe2y2nmSRkONzuVuO50uPtSCh1WyDAnuqHiIee4YYCjtZ1Vwo1WrOiLjCZANsNGqZLLOvNdOyC33Q3hYAE8IKcLtdbsHDlH7pjlTD0XMQa6CcRMARlgxDhpQ06MVsS4JDgA3MTSrh%2B8ssKsQOYj0vT2eSBcV6swzOJ5Pq7MojMNiOJg1GkD52ja52rx0cBZGjgAVl4vg4WlIqE4risVgfEsKyYA8kjIjwpAEJoT4LAgmBMFg0QrqQADWIDIjEAB0XCvhoABs%2BEABwxDEyL4RoGgxAALPonDUbwLASJRpCft%2Bv4cLwCYaFBMELHAsAwIgKCoCw8R0FE5CUGgYkSdEwBcIpfB0AQUQJhA4QwaQ4RBLUlacJBOnMMQlZXOE2gVNB3C8DJbCCFcDC0PpX68Fg4SeMArgUgm1mkFgLCGMA4guX5eC9pUU6YD536YKoFSeKpBm8KWrRabQeDhMQenuOqvlEngTG%2BZFxDhEkmBopgAVGOlRh8XwBjAAoABqeCYLKNyMElMiCCIYjsFwPH8IIigqOoIW6M0Bi1aYFhWPoGUJpACxRncnC8KgxXEMKi0oeUlROBALgjE0pABFMhTFFkSQpAIx1XTkqQ9BdswtG0VQTHdYytJZ71dE9fQlIMXSfUDdT/TMJQLAowGrBIz6cO%2BrFaRxhyqER%2BEALT4dRhzAMgyCHIpOGHBAuCECQYEQXMvBWVo/FIDJ4n0GQFAQIzckgApSk0LQqnEOpmkhUZeldcLJlmRZ9hdbZjAEA5TlaW5HleYaPmQf5gXBd%2B%2BDhQ4kXRbwsXxYlvkpS%2B37pZl2UYGskH5YVkHFaVSgVVVQVBKALkLFQDXNa17XxJ1vlDcIojiAN3XyEoahabo0hTcYAEWPN4Q7ctZ7xmtP6bdt8BQ99%2B2%2BIdDBuB4jR%2BCX4OXc02Q3WkZejAk13tFXL17e0nTDA3J3t79YPnQDejjMD3dDxMreQ4syyw1w8Nvh%2ByOcKj6NYzjeME0T86k/gRDEJTs803xCzoXhdEcAxpBMafbHrVnCbNLTT6kAJwlLAQ8QJVJh07yQQJ6CHvVxBEUjiNGO409CyiyvEOqL5EY3x/JwK4CUP7fFQFQZemNsa43xoTLgxMIDuFkszfe1NeJezgghJClA54XyvjxeBHEuIgB4o/OYx8QCSEeFhV8uFSKPHwo8ZEBFXyCORGfZEC8QqMLIXTZ%2BQkIAMxtmTXeX8eaMFYLbXgACw79VoiHUBY1vxx30IFFASdrC9wOkdUezQzoFEHqUJuD1bo2KcXXCeQ8C4dw%2Bq4yxHRx4DwhgMYeXcMg9wCfYoJs8p4gU4AAejiYcUklZA6HErMBQ4aBgTwV7IiCmtBUCIUOIYQ4sVWDxHoFhCc3wFDik8LUTAEB1z0k7O2fkbxez9mIIOekhwABUhx6Qji5OOLYnwjhtM2AScZ3xBRqlAjCUpvSTRdjTIie0RZNntMJMSWZ%2BoKRdhpHSaEzwbysg2c6HUlzHQEjmSmUUvTJTIBlPKLAI5bn7NzA8iUxBzQUmGc6b0GcFCbF5lwYZYyiRfC7LqN4Ry7nAhhGstknJrndnePCz5lJGTMlvOmFFWypl6kPKBJ5LzhTvL1JeQ4ZoLS0ABcqTMoKCCSBHE/WBkj2J/nMUBaeCzwKSEPuQtCGFkRYWRBKyVUqpVnwvgwu%2BzCZGwTPoKpGUis6sIWMVZIThqJAA%3D

TeachRaccooon: And it becomes clearer because the user knows that we're using an enumerations?
Thats it?

rileyjmurray: What would your version of the following signature look like if we used your preferred method?
```c++
template <typename T>
void bqrrp_functional(
    int64_t m, int64_t n, T* A, int64_t lda, T* tau, int64_t* piv,
    Options::QRTall::Enum      qr_tall       = Options::QRTall::QRF,
    Options::QRCPWide::Enum    qrcp_wide     = Options::QRCPWide::LUQR, 
    Options::ApplyTransQ::Enum apply_trans_q = Options::ApplyTransQ::ORMQR
);
```

TeachRaccooon: Also, idk if there's any sense in making a separate function that allows for tuning optional stuff, since it is optional.
The user may want to choose one of the subroutine types to be changed from the default, while not willing to think at all about the other subroutine types.

rileyjmurray: The plan for RandLAPACK has always been to have an object-oriented API and a procedural API. I just want to know what the procedural API would look like with your approach.

TeachRaccooon: Wouldnt it just be

```
template <typename T>
void bqrrp_procedural_api(
    int64_t m, int64_t n, T* A, int64_t lda, T* tau, int64_t* piv,
    Options::QRTall      qr_tall       = Options::QRTall::QRF,
    Options::QRCPWide    qrcp_wide     = Options::QRCPWide::LUQR, 
    Options::ApplyTransQ apply_trans_q = Options::ApplyTransQ::ORMQR
);
```

if i were to change names?

rileyjmurray: Please verify on Godbolt.

TeachRaccooon: Idk how that platform works, but i just added this into one of RandLAPACK tests and it compiles just fine.
The syntax is the same as what I've pointed out, but the names are obviously different.

```
...
    static void buffer1(
        RandLAPACK::BQRRP<T, RNG>::SubroutineType::qr_tall_subroutines qr_tall) {
            int a = 0;
        }

    static void buffer2(
        RandLAPACK::BQRRP<T, RNG>::SubroutineType::qr_tall_subroutines qr_tall = RandLAPACK::BQRRP<T, RNG>::SubroutineType::qr_tall_subroutines::cholqr) {
            int a = 0;
        }
};

TEST_F(TestBQRRP, test_buf) {
    RandLAPACK::BQRRP<double, r123::Philox4x32>::SubroutineType::qr_tall_subroutines qr_tall = RandLAPACK::BQRRP<double, r123::Philox4x32>::SubroutineType::qr_tall_subroutines::cholqr;
    buffer1<double, r123::Philox4x32>(qr_tall);
    buffer2<double, r123::Philox4x32>();
}
```

rileyjmurray: Looks like the compiler is indeed happy with this
```c++
struct Options {
    enum QRCPWide {LUQR, QP3};
    enum QRTall {QRT, CholQR, QRF};
    enum ApplyTransQ {ORMQR, GEMQRT};
};

// below is just a demo.
template <typename T>
int bqrrp_procedural_api(
    int m, int n, T* A, int lda, T* tau, int* piv,
    Options::QRTall      qr_tall       = Options::QRTall::QRF,
    Options::QRCPWide    qrcp_wide     = Options::QRCPWide::LUQR, 
    Options::ApplyTransQ apply_trans_q = Options::ApplyTransQ::ORMQR
) {
    return 0;
}
```
Suggestion: define the struct type outside the BQRRPalg class and name it BQRRPOptions. Then at the top of the BQRRPalg class have a line
```c++
using Options = BQRRPOptions;
```
If you take this approach then the BQRRP implementation can use the shorter Options symbol, while test code can refer to BQRRPOptions like ``RandLAPACK::BQRRPOptions::QRTall::CholQR`` instead of ``RandLAPACK::BQRRP<double, r123::Philox4x32>::Options::QRTall::CholQR``.

TeachRaccooon: Yep, which is the same as what I've shown, just different names. 

Can do, I just thought that we wanted this struct to be formally attached to the BQRRP class.
Then again, since the struct was public, that doesn't make too much difference. 
Will apply this change a bit later.

TeachRaccooon: @rileyjmurray wait a sec, your initial complain about the way the enums were set was related to the fact that my enums pollute the RandLAPACK namespace and now you're suggesting we throw them away from the BQRRP class, which, technically, also pollutes the RandLAPACK namespace.

rileyjmurray: You're right to ask this question.

In an ideal world the enum(s) would live inside BQRRP. The problem is that since BQRRP is templated we can't just write ``BQRRP::Options::QRTall::CholQR``. Even if we used a default template parameter for the RNG we'd have to specify the template parameter for the numerical precision. That's awkward because the enum has no semantic tie to the numerical precision.

Having a single BQRRPOptions struct -- which contains all enums inside of it -- lets us group these options together without needing to specify unrelated template parameters of a BQRRP class.

TeachRaccooon: Got it, just pushed the change to that effect; the names are different because "Subroutines" instead of "Options" matches the paper contents, and the function names are lowercase full function names because that is cleaner for the user and I use lowercase in other enums in RandLAPACK.

rileyjmurray (approved):

PR #97, opened 2025-02-01
Replace class-inheritance-based APIs for symmetric matrix algorithms with abstract templating APIs

rileyjmurray:
Changes
* Defined a SymmetricLinearOperator C++20 concept. It only requires that conforming types have an int64_t "dim" member and that operator() is implemented with a certain signature. Removed the base classes for ``ExplicitSymLinOp`` and ``RegExplicitSymLinOp``.
* Remove base classes for SYPS, SYRF, and REVD2. The purposes of these base classes can be handled in a simpler way by changing how we template REVD2 and SYRF.
* Adopt a new naming convention for members of algorithm classes. This is best exemplified with SYRF. It takes two template parameters: ``SYPS_t`` and ``Orth_t``, which are the types of instance variables called ``syps`` and ``orth`` --- note the lower case. At present, the only valid choice for ``SYPS_t`` is ``RandLAPACK::SYPS<T,RNG>`` for some scalar type ``T`` and CBRNG type ``RNG``. The suffix "_t" in ``SYPS_t`` helps distinguish between the abstract template parameter with our available implementation in the templated ``RandLAPACK::SYPS`` type.
* I've moved some function implementations inside the original class definition. (Before the class definition only contained function signatures. That style is unnecessary given that RandLAPACK has implementations in the header files.)

### DIFF
--- a/RandLAPACK/comps/rl_determiter.hh
+++ b/RandLAPACK/comps/rl_determiter.hh
@@ -368,7 +368,7 @@ int64_t posm_square(
  *      A boolean. If true, then logging information is printed to stdout.
  * 
  */
-template <typename T, linops::SymmetricLinearOperator FG, typename FSeminorm, typename FN>
+template <typename T, typename FG, typename FSeminorm, typename FN>
 void pcg(
     FG &G,
     const T* H,
@@ -510,7 +510,7 @@ void pcg(
  *       
  * The main purpose of this wrapper is backward-compatibility with existing code.
  */
-template <typename T, linops::SymmetricLinearOperator FG, typename FN, typename FSeminorm = StatefulFrobeniusNorm<T>>
+template <typename T, typename FG, typename FN, typename FSeminorm = StatefulFrobeniusNorm<T>>
 FSeminorm pcg(
     FG &G,
     const std::vector<T> &H,
--- a/RandLAPACK/comps/rl_determiter.hh
+++ b/RandLAPACK/comps/rl_determiter.hh
@@ -290,7 +290,7 @@ int64_t posm_square(
  *  G_i * x_i = h_i for i in {1, ..., s}        (Eq. 1)
  *
  * with preconditioners N_1, ..., N_s. The linear systems all have the same number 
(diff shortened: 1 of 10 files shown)


rileyjmurray:
Here's a note to future contributors (possibly myself...) about why we can't use C++20 concepts to enforce constraints equivalent to those of an abstract class. My response to this was just to use typenames like ``SYPS_t`` in unconstrained templates, and just let the compiler figure things out. 
```c++
/*
Looks like there's no way to define a concept for a single type, SYPS_t,
where SYPS_t satisftying the SymmetricPowerSketcherConcept implies that 
objects of type SYPS_t have a call(...) method whose first argument is
templated to accept any object whose type satisfies SymmetricLinearOperator.

I can only define a concept that accepts two template parameters. Those
can, in principle, be used for templating with patterns of the form

    template <typename SYPS_t, typename SLO>
    requires SymmetricPowerSketchConcept<SYPS_t, SLO>
    // ... definition here ...

*/

// In the concept below the template parameters T and RNG are just aliases for readibility.
// Because they have default values, an expression "SymmetricPowerSketchConcept<SYPS_t, SLO>""
// is a well-formed way of enforcing the SymmetricPowerSketchConcept requirement on (SYPS_t, SLO).
template <typename SYPS_t, typename SLO, typename T = SYPS_t::scalar_t, typename RNG = SYPS_t::RNG_t>
concept SymmetricPowerSketchConcept = 
    linops::SymmetricLinearOperator<SLO> &&
    requires(SYPS_t obj, Uplo uplo, int64_t m, const T* A, int64_t lda, int64_t k, RandBLAS::RNGState<RNG> &state, T* skop_buff, T* work_buff) {
        { obj.call(uplo, m, A, lda, k, state, skop_buff, work_buff) } -> std::same_as<int>;
        { obj.call(std::declval<SLO>(), k, state, skop_buff, work_buff) } -> std::same_as<int>;
};
```

rileyjmurray:
@TeachRaccooon this is ready for review.

Issue #75, opened 2024-09-07
Printf vs cout.


TeachRaccooon:
We need to substitute printf with cout all throughout the codebase.

PR #114, opened 2025-12-02
Print overhaul

mmelnich:
Converted 150+ printf statements to std::cout across library, tests, and benchmarks (not using streams where applicable yet).

Issue #75.


mmelnich:
@rileyjmurray let me know what you think

rileyjmurray (approved):

PR #115, opened 2025-12-02
Winter 2025 linops 

mmelnich:
WIP

PR #119, opened 2026-01-14
Jan 2025 linops

mmelnich:
This PR is a part of of #115, which will be broken up into more digestible parts and closed.
This PR intoduces refined abstract linear operator support (with testing) and adds a new "composite" linear operator type.

Core changes:

1. Linear Operators (rl_linops.hh):
    -   Extended DenseLinOp and SparseLinOp with full multiplication support (both Side::Left and Side::Right)
    -   Added row-major layout support for dense operators
    -   Added CompositeLinOp class for operator composition (e.g., A = B * C)
    -   Added tests for each linop type

2. Matrix Generation (rl_gen.hh)
    - Added gen_random_dense() with layout support
    - Added gen_sparse_coo() and gen_sparse_csc() for sparse matrix generation
    - New matrix generation functions do not rely on "mat_info" struct so that the matrices can be generated straightforwardly

3. Utilities (rl_util.hh, rl_util_test.hh)
    - Reorganized utility functions used specifically in tests into separate header (rl_util_test.hh)
    - Added sparse_to_dense_summing_duplicates() for sparse-to-dense conversion

4. Drivers
    - Minor updates to rl_abrik and its benchmarks for linop compatability (unsure why the whole driver file appears reworked)
    - rl_cqrrt is refactored to make sure it only processes full-rank matrices

### DIFF
--- a/test/misc/test_linop_unified.cc
+++ b/test/misc/test_linop_unified.cc
@@ -145,8 +145,8 @@ RandLAPACK::linops::SparseLinOp<RandBLAS::sparse_data::csc::CSCMatrix<T>> make_o
 ) {
     data.rows = rows;
     data.cols = cols;
-    // Use emplace with move since CSCMatrix has deleted copy assignment
-    data.csc_mat.emplace(std::move(RandLAPACK::gen::gen_sparse_csc<T>(rows, cols, density, state)));
+    // Use emplace since gen_sparse_csc returns by value (already an rvalue)
+    data.csc_mat.emplace(RandLAPACK::gen::gen_sparse_csc<T>(rows, cols, density, state));
 
     return RandLAPACK::linops::SparseLinOp<RandBLAS::sparse_data::csc::CSCMatrix<T>>(
         rows, cols, *data.csc_mat);
@@ -197,7 +197,7 @@ void densify_operator(
 // Composite operator creation and densification
 // ============================================================================
 
-// Composite operator creation - creates left and right operators and returns composite
+// Composite operator creation
 // IMPORTANT: The operators are stored in data.left_op and data.right_op because
 // CompositeOperator stores references, so we need the operators to outlive the CompositeOperator.
 template <typename LeftTag, typename RightTag, typename T = typename LeftTag::scalar_t>
@@ -308,8 +308,10 @@ protected:
     ///    - Compute C_reference using BLAS gemm directly on dense matrices.
     ///
     /// 5. VERIFICATION
-    ///    - Compare C_op and C_reference entry-wise.
-    ///    - Uses tolerance: atol = 100 * eps, rtol = 10 * eps.
+    ///    - Compare C_op and C_reference entry-wise using componentwise error bounds.
+    ///    - Error bound: |C_op - C_ref| <= |alpha| * k * 2 * eps * |A| * |B| + |beta| * eps * |C_old|
+    ///    - This bound accounts for rounding errors in GEMM proportional to the inner dimension
+    ///      and the magnitudes of the operands.
     ///
     /// Parameters:
     ///    test_linear_operator<OpTag>(side, sparse_B, layout, trans_A, trans_B, m, n, k, params)
@@ -391,6 +393,12 @@ private:
         T* C_reference = new T[m * n];
         std::copy(C_op, C_op + m * n, C_reference);
 
+        // Store |C_old| for error bound computation (beta term)
+        T* E = new T[m * n];  // Error bound matrix
+        for (int64_t i = 0; i < m * n; ++i) {
+            E[i] = std::abs(C_op[i]);
+        }
+
         // Allocate dense matrices for reference computation
         T* A_dense = new T[dims.rows_A * dims.cols_A]();
         T* B_dense = new T[dims.rows_B * dims.cols_B]();
@@ -420,20 +428,43 @@ private:
                                A_dense, dims.lda, B_dense, dims.ldb,
                                beta, C_reference, dims.ldc);
 
-        // Compare results
-        T atol = 100 * std::numeric_limits<T>::epsilon();
-        T rtol = 10 * std::numeric_limits<T>::epsilon();
-        test::comparison::matrices_approx_equal(
-            layout, Op::NoTrans, m, n, C_op, dims.ldc,
-            C_reference, dims.ldc, __PRETTY_FUNCTION__, __FILE__, __LINE__,
-            atol, rtol
+        // Compute componentwise error bound matrix E
+        // Error bound: |C_computed - C_exact| <= |alpha| * inner_dim * 2 * eps * |A| * |B| + |beta| * eps * |C_old|
+        // We use GEMM on absolute values to compute this bound
+        T eps = std::numeric_limits<T>::epsilon();
+        T err_alpha = std::abs(alpha) * k * 2 * eps;
+        T err_beta = std::abs(beta) * eps;
+
+        // Compute |A| and |B|
+        T* A_abs = new T[dims.rows_A * dims.cols_A];
+        T* B_abs = new T[dims.rows_B * dims.cols_B];
+        for (int64_t i = 0; i < dims.rows_A * dims.cols_A; ++i) {
+            A_abs[i] = std::abs(A_dense[i]);
+        }
+        for (int64_t i = 0; i < dims.rows_B * dims.cols_B; ++i) {
+            B_abs[i] = std::abs(B_dense[i]);
+        }
+
+        // E = err_alpha * |A| * |B| + err_beta * |C_old|
+        // (E was initialized with |C_old|)
+        compute_gemm_reference(side, layout, trans_A, trans_B, m, n, k, err_alpha,
+                               A_abs, dims.lda, B_abs, dims.ldb,
+                               err_beta, E, dims.ldc);
+
+        // Compare results using componentwise bounds
+        test::comparison::buffs_approx_equal(
+            C_op, C_reference, E, m * n,
+            __PRETTY_FUNCTION__, __FILE__, __LINE__
         );
 
         // Clean up
         delete[] A_dense;
         delete[] B_dense;
+        delete[] A_abs;
+        delete[] B_abs;
         delete[] C_op;
         delete[] C_reference;
+        delete[] E;
     }
 };
 


RandLAPACK/misc/rl_gen.hh
```diff
@@ -522,8 +647,8 @@ void mat_gen(
 }
 
 template <typename T, typename RNG>
-std::vector<T> mat_gen(mat_gen_info<T> &info, RandBLAS::RNGState<RNG> &state) {
-    std::vector<T> A(info.rows * info.cols, 0.0);
+::std::vector<T> mat_gen(mat_gen_info<T> &info, RandBLAS::RNGState<RNG> &state) {
```
rileyjmurray: What is this leading double colon syntax for?

RandLAPACK/misc/rl_linops.hh
```diff
@@ -27,195 +32,978 @@ concept LinearOperator = requires(LinOp A) {
...
+
+            if (trans_comp == Op::NoTrans) {
+                // C := alpha * (LinOp1 * LinOp2) * op(B) + beta * C
+                // Intermediate dimension: right_op.n_rows = left_op.n_cols
+                int64_t temp_rows = right_op.n_rows;
+                T* temp_buffer = new T[temp_rows * n]();
+                int64_t ldt = (layout == Layout::ColMajor) ? temp_rows : n;
+
+                // Step 1: temp := LinOp2 * op(B), dimension (right_op.n_rows x n)
+                right_op(Side::Left, layout, Op::NoTrans, trans_B, temp_rows, n, k, (T)1.0, const_cast<T*>(B), ldb, (T)0.0, temp_buffer, ldt);
```
rileyjmurray: Why the const cast?

RandLAPACK/misc/rl_util.hh
```diff
@@ -490,4 +376,168 @@ std::string get_current_date_time() {
...
+    return false;
+}
+
+/// Computes an L-2 norm of a given matrix using
+/// p steps of power iteration.
+template <typename T, typename RNG>
+T estimate_spectral_norm(
+    int64_t m,
+    int64_t n,
+    T const* A_dat,
```
rileyjmurray: I suggest you implement this function using abstract LinearOperator objects. You can have a version that accepts BLAS-style matrices (declared layout and LDA) by wrapping the LinearOperator implementation.

RandLAPACK/misc/rl_util.hh
```diff
@@ -490,4 +376,168 @@ std::string get_current_date_time() {
...
+    T prev_norm_inv = 1.0;
+    for(int i = 0; i < p; ++i) {
+        // A * v
+        gemv(Layout::ColMajor, Op::NoTrans, m, n, 1.0, A_dat, m, buf.data(), 1, 0.0, buf1.data(), 1);
+        // prev_norm_inv * A' * A * v
+        gemv(Layout::ColMajor, Op::Trans, m, n, prev_norm_inv, A_dat, m, buf1.data(), 1, 0.0, buf.data(), 1);
+        prev_norm_inv = 1 / blas::nrm2(n, buf.data(), 1);
+    }
+
+    return std::sqrt(blas::nrm2(n, buf.data(), 1));
```
rileyjmurray: This looks weird. Leave a comment explaining what’s happening.

RandLAPACK/misc/rl_util.hh
```diff
@@ -490,4 +376,168 @@ std::string get_current_date_time() {
...
+}
+
+/// Convert sparse matrix to dense format, summing duplicate entries if present.
```
rileyjmurray: I’m pretty sure RandBLAS already has conversion functions for this.

RandLAPACK/misc/rl_util_test.hh
rileyjmurray: Tons of this leading ``::`` syntax. Why?

rileyjmurray (changes requested):
There are several places where constants like 0.0 or 1.0 are not cast to the template parameter type T before invoking a BLAS function. This can result in dispatching BLAS++ implementations that are inefficient. Please open a GitHub issue about this. Once you’re done with the current batch of PRs you can open a new PR to address this. It should be easy to catch most offenses; just CTRL+F for “, 0.0” and “, 1.0”.

I’ve reviewed all files up until the ``benchmark/`` folder.

test/misc/test_linop_composite.cc
rileyjmurray: Please use appropriate utility functions for changes-of-layout. There are a few places where you write out the loops explicitly, which is undesirable.

RandLAPACK/misc/rl_util.hh
mmelnich: I can, but this function is not used anywhere at all right now. 
I wrote it back when we were looking for ways to estimate rank in CQRRPT.

test/misc/test_linop_composite.cc
rileyjmurray: I want to see comments that describe, in mathematical terms, how TestCompositeLinOp::test_composite_linop is structured. Like, “We generate buffer representations of A, B, C, where A and B might contain lots of zeros. Then we construct appropriate linear operator wrappers of A and B. In some cases the linear operator is SparseOperator, while in others it’s a DenseOperator. SparseOperator objects are backed by CSCMatrix objects as needed. If side=right, then we compute D = 1.5 op(A) op(B) C+ 0.5 * C using the composite operator object for (A, B) and using GEMM. Then we check for whenever the equality up to tolerance [TOL].”

^Revise to make that correct and expand it.

RandLAPACK/misc/rl_util.hh
mmelnich: Yet another ancient function.
Comments added, nontheless.

test/misc/test_linop_sparse.cc
rileyjmurray: I suspect you can use templating to write tests that apply to anything that implements the LinearOperator concept. There’s plenty of that goes into preparing the data that would be provided to such a templated test, but you still save complexity overall if you go the templated route and merge tests for dense and sparse linops. Think of it this way: I want minimal code duplication if we have to write tests for yet another class of linear operators. I’ll have to open a PR soon anyway and add support for a cholesky-solve linear operator!

rileyjmurray (changes requested):
This finishes the review of your initial 9 commits.

RandLAPACK/misc/rl_gen.hh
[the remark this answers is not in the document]
mmelnich: Ok yeah you're right, there was an issue.
Also, there was no "deduplication" like there is in RandBLAS.
Thats why using RandBLAS's to_dense gave bad results.

RandLAPACK/misc/rl_util_test.hh
```diff
@@ -0,0 +1,147 @@
...
+}
+
+// ============================================================================
+// Layout Conversion Helper
+// ============================================================================
+
+/// Convert matrix layout in-place
+/// Note: This assumes the matrix is currently in 'from_layout' and converts to 'to_layout'
+template <typename T>
+void convert_layout_inplace(
```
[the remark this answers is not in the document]
mmelnich: Ok yeah you're right.

test/drivers/test_abrik.cc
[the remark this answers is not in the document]
mmelnich: Removed them for now.

RandLAPACK/misc/rl_util.hh
```diff
@@ -490,4 +376,179 @@ std::string get_current_date_time() {
...
+    const T* A,
+    bool verbose
+) {
+    T* A_cpy = new T[m * n]();
+    T* s     = new T[n]();
+
+    lapack::lacpy(MatrixType::General, m, n, A, m, A_cpy, m);
+    lapack::gesdd(Job::NoVec, m, n, A_cpy, m, s, NULL, m, NULL, n);
+
+    T cond_num = s[0] / s[n - 1];
```
rileyjmurray: if ``s[n-1]`` is zero then we should make sure to return an IEEE-compliant inf. NaNs are bad.

RandLAPACK/misc/rl_gen.hh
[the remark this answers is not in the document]
rileyjmurray: Consider opening an issue in the RandBLAS repo pointing this out. Right now the sparse tests in RandBLAS construct sparse matrices by iid-sparsifying random dense matrices. That precludes the possibility of duplicates.

RandLAPACK/misc/rl_util_test.hh
mmelnich: Sure.

RandLAPACK/misc/rl_util.hh
mmelnich: Done.

test/misc/test_linop_sparse.cc
mmelnich: Done.

test/misc/test_linop_unified.cc
```diff
@@ -0,0 +1,922 @@
...
+    std::fill(dense_out, dense_out + data.rows * data.cols, T(0));
+    RandBLAS::sparse_data::csc::csc_to_dense(*data.csc_mat, target_layout, dense_out);
+}
+
+// ============================================================================
+// Composite operator creation and densification
+// ============================================================================
+
+// Composite operator creation - creates left and right operators and returns composite
+// IMPORTANT: The operators are stored in data.left_op and data.right_op because
```
rileyjmurray: I like this note.

test/misc/test_linop_unified.cc
```diff
@@ -0,0 +1,922 @@
...
+    ///        Side::Right: C_op := alpha * op(B) * op(A) + beta * C_op
+    ///
+    /// 4. REFERENCE COMPUTATION
+    ///    - Convert A to dense format using densify_operator<OpTag>.
+    ///    - If B is sparse, convert it to dense format as well.
+    ///    - Compute C_reference using BLAS gemm directly on dense matrices.
+    ///
+    /// 5. VERIFICATION
+    ///    - Compare C_op and C_reference entry-wise.
+    ///    - Uses tolerance: atol = 100 * eps, rtol = 10 * eps.
```
rileyjmurray: The factors 100 and 10 aren't appropriate in general. You should use the error bound guaranteed by non-Strassen GEMM (e.g., naive three loop GEMM, or tiled GEMM). The componentwise bound depends on the entries of the operands to the matrix-matrix product.

Take a look at the following RandBLAS testing function. It uses GEMM to compute $C$ from densified $(A,B)$ __and__ an elementwise nonnegative matrix $E$ (depending on working precision, epsilon) so that our test becomes $|C_ {\text{linop}} - C| \leq E$ componentwise.

https://github.com/BallisticLA/RandBLAS/blob/4d2f7ef8d45d292fee61a668931bddaf87fee9cd/test/test_matmul_cores/linop_common.hh#L197-L268

For your purposes you can assume ``S_ro == S_co == 0``.

rileyjmurray (changes requested):
See my note about error bounds.

test/misc/test_linop_unified.cc
```diff
@@ -0,0 +1,922 @@
...
+    int64_t rows,
+    int64_t cols,
+    Layout layout,  // layout doesn't affect CSC storage
+    T density,
+    RNGState<r123::Philox4x32_R<10>>& state
+) {
+    data.rows = rows;
+    data.cols = cols;
+    // Use emplace with move since CSCMatrix has deleted copy assignment
+    data.csc_mat.emplace(std::move(RandLAPACK::gen::gen_sparse_csc<T>(rows, cols, density, state)));
```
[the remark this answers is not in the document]
rileyjmurray: Cool, then please remove.

test/misc/test_linop_unified.cc
mmelnich: So just to double-check, you're saying that if we're computing `C = alpha * A * B + beta * C`, then the correct expression would be the following (elementwise):
`|C_computed - C_exact(via ref gemm)| < 2 * m * eps * |alpha| * |A| * |B| + |beta| * eps * |B|`?

mmelnich: Just pushed the change

rileyjmurray (approved):

PR #130, opened 2026-03-27
ABRIK Wrap-Up

mmelnich:
WIP


rileyjmurray:
I investigated the failures at `608ad40` and pushed fixes in [3c4d68e](https://github.com/BallisticLA/RandLAPACK/commit/3c4d68e30fe62a870e1e8136f4bac684b58b4ab9), together with the main update containing #181.

The ten failing jobs reduce to six BK/ABRIK test assertions repeated across platforms. Compilation succeeded; the ASAN jobs also failed on assertions, with no sanitizer finding in those logs. Three adaptive failures were already present in the [August 12 Linux run](https://github.com/BallisticLA/RandLAPACK/actions/runs/31646746957). The BK/ABRIK implementations and their tests, and the RandBLAS pin, are unchanged between the premerge `0ccab55` and `608ad40`. An earlier merge had added Windows coverage and separate ASAN jobs, so more jobs now expose the same problems. macOS also now uses the new Accelerate interface, which can change numerical paths near a threshold. The extraction merge did not newly enable these tests.

The fixes are:

- **Resume after narrowing:** the iteration-budget check masked a terminal norm-convergence decision. At iteration 5, BK could report resumable even though it had finished; resuming performed an unnecessary even iteration. Terminal norm/saturation checks now precede the budget check. A small new regression fails before this fix and passes afterward. The narrowed-resume test checkpoints at iteration 4, after narrowing but before convergence, and still compares the completed bases exactly.
- **Even CQRRT band identity:** the test assumed a longer run would end with an even zero-width rank probe. Some backends correctly stopped one step earlier through norm convergence, with a valid band. The test now explicitly stops at the narrowed even step and verifies the intended 25-by-20 band, width, and reconstruction identity.
- **Three adaptive accuracy checks:** these retained an absolute-error threshold that had already been replaced in the shared test harness. One also assessed all 100 returned triplets although the driver was asked to assess the leading 20. The tests now independently calculate the normalized residual, check the requested rank and convergence status, and use the existing all-triplet quality criteria for fallback termination. The adaptive/non-adaptive comparison also checks leading singular values.
- **Early stopping on a decaying spectrum:** `1e-14` was close enough to the numerical floor that some backends or seeds reached the full subspace. The test now requests `1e-10`, still requires convergence before saturation and a directly measured residual below that tolerance, and checks that fewer than all columns were returned. Across three seeds it stopped after 16 iterations. Restoring the old all-triplet assessment makes the revised regression fail at full-subspace termination.

Local validation: all **416 core tests** and **142 extras tests** pass, and all four ABRIK benchmark callers compile with float and double. **Cross-platform CI on `3c4d68e` now passes as well:** all 11 jobs, including Linux/macOS ASAN, Windows OpenBLAS and MKL, and all three installer jobs, plus the CLA check.

--- Astra

rileyjmurray:
The fixes above address every assertion failure observed in the original CI logs. They do leave two numerical questions for human review; I kept these out of the straightforward CI repair.

1. **Accuracy on terminal fallback.** I corrected stale test assertions, but did not improve the accuracy attained when BK reports norm convergence, rank deficiency, or saturation before ABRIK's requested residual tolerance is met. The [norm threshold](https://github.com/BallisticLA/RandLAPACK/blob/3c4d68e30fe62a870e1e8136f4bac684b58b4ab9/RandLAPACK/comps/rl_bk.hh#L571) uses `sqrt(1 - tol^2) * ||A||_F`; sufficiently small `tol^2` is lost in floating-point subtraction. That content test cannot resolve an arbitrarily small residual tolerance. The existing [driver fallback](https://github.com/BallisticLA/RandLAPACK/blob/3c4d68e30fe62a870e1e8136f4bac684b58b4ab9/RandLAPACK/drivers/rl_abrik.hh#L426) reports a distinct terminal reason. The numerical-method review should decide whether that behavior is the intended contract, or whether refinement or a more reliable stopping estimate is needed. Changing the CI tests does not make the previous near-machine-precision accuracy demand attainable.

2. **Termination near numerical rank boundaries.** I fixed the budget/status precedence bug and made the even-band test exercise its intended geometry. I did not force every BLAS/LAPACK backend to choose the same norm-convergence versus zero-width-rank-probe exit. For the rank-25 fixture, both routes produced a valid band, while tiny differences near the norm/rank thresholds changed the final parity. Whether the final BK interface should guarantee reproducible termination classifications, and how its [rank threshold](https://github.com/BallisticLA/RandLAPACK/blob/3c4d68e30fe62a870e1e8136f4bac684b58b4ab9/RandLAPACK/comps/rl_bk.hh#L567) should interact with content convergence, warrants deliberate numerical analysis rather than another CI-driven threshold change.

No failing test was disabled. All CI checks on `3c4d68e` now pass. The numerical-method questions above remain open for human review.

--- Astra

PR #136, opened 2026-05-10
fix compiler warnings under -Wall -Wextra

mmelnich:
**Silence all RandLAPACK compiler warnings under `-Wall -Wextra`**

Resolves all 22 warnings emitted when building the workshop docker image (`ghcr.io/rileyjmurray/randlapack-python-workshop:latest`).

**Categories fixed** (-Wunused-but-set-variable, -Wunused-parameter, -Wsign-compare, -Wvolatile, -Wdangling-else):
- Removed dead code: `prevnormNR` in `rl_determiter.hh`; unused `[rows_B, cols_B]` structured bindings in `rl_dense_linop.hh` / `rl_composite_linop.hh`.
- Removed genuinely vestigial parameters: `lda` from sparse `ABRIK::call` (LDA is meaningless for sparse — `test_ABRIK_general` uses `if constexpr` to dispatch dense vs. sparse), `k` from `test_csr_col_split`, `norm_A` from `test_CQRRPT_orthogonalization`. All callers updated.
- For parameters that are unused only in one specialization but required by template-overload symmetry (`density`/`layout` in `make_operator`), used the existing `UNUSED(x)` macro from `RandBLAS/exceptions.hh` with explanatory comments.
- Misc: braced `ASSERT_FALSE` in `test_rpchol.cc` for `-Wdangling-else`; rewrote `x += i` on `volatile int` as `x = x + i` to avoid C++20 deprecation in `test_memory_tracker.cc`.

Added issue #137 to deal with clang warnings/errors (will involve updating RandBLAS submodule).

### DIFF
--- a/RandLAPACK/drivers/rl_abrik.hh
+++ b/RandLAPACK/drivers/rl_abrik.hh
@@ -139,7 +139,7 @@ class ABRIK {
             int64_t m,
             int64_t n,
             SpMat &A,
-            int64_t lda,
+            [[maybe_unused]] int64_t lda,
             int64_t k,
             T* &U,
             T* &V,
--- a/RandLAPACK/drivers/rl_abrik.hh
+++ b/RandLAPACK/drivers/rl_abrik.hh
@@ -139,7 +139,6 @@ class ABRIK {
             int64_t m,
             int64_t n,
             SpMat &A,
-            [[maybe_unused]] int64_t lda,
             int64_t k,
             T* &U,
             T* &V,
--- a/test/linops/test_linop_block_views.cc
+++ b/test/linops/test_linop_block_views.cc
@@ -786,7 +786,7 @@ protected:
 
     /// Test csr_split_col_blocks: split and verify all blocks are correct.
     template <typename T>
-    void test_csr_col_split(int64_t m, int64_t n, [[maybe_unused]] int64_t k, T density, int64_t num_blocks) {
+    void test_csr_col_split(int64_t m, int64_t n, T density, int64_t num_blocks) {
         RNGState state(42);
         auto A = generate_csr_matrix<T>(m, n, density, state);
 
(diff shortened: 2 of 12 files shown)


test/linops/test_linop_block_views.cc
```diff
@@ -786,7 +786,7 @@ class TestCSRColBlocks : public ::testing::Test {
 
     /// Test csr_split_col_blocks: split and verify all blocks are correct.
     template <typename T>
-    void test_csr_col_split(int64_t m, int64_t n, int64_t k, T density, int64_t num_blocks) {
+    void test_csr_col_split(int64_t m, int64_t n, [[maybe_unused]] int64_t k, T density, int64_t num_blocks) {
```
rileyjmurray: I just don't see k used here anywhere at all

test/linops/test_linop_unified.cc
```diff
@@ -140,7 +140,7 @@ RandLAPACK::linops::SparseLinOp<RandBLAS::sparse_data::csc::CSCMatrix<T>> make_o
     OperatorData<SparseCSCOpTag<T>>& data,
     int64_t rows,
     int64_t cols,
-    Layout layout,  // layout doesn't affect CSC storage
+    [[maybe_unused]] Layout layout,  // layout doesn't affect CSC storage
```
rileyjmurray: There's no "maybe" here. The function body is very short and clearly doesn't reference "layout".

RandLAPACK/drivers/rl_abrik.hh
```diff
@@ -139,7 +139,7 @@ class ABRIK {
             int64_t m,
             int64_t n,
             SpMat &A,
-            int64_t lda,
+            [[maybe_unused]] int64_t lda,
```
rileyjmurray: Why would we ever use LDA for a sparse matrix?

test/linops/test_linop_block_views.cc
mmelnich: Fixed

RandLAPACK/drivers/rl_abrik.hh
mmelnich: Fixed.

test/linops/test_linop_unified.cc
mmelnich: Robot's pushback:
The CSC overload doesn't use layout and the dense overload doesn't use density, but both signatures must match: make_operator is selected by OperatorData<TAG> type and templated callers (e.g. line 364) call it uniformly. [[maybe_unused]] is exactly that case — required by the overload contract, unused in this specialization.

Happy to refactor (split into named functions + if constexpr, or an OperatorParams struct) if you'd prefer, but the current overload set needs the symmetric signature.

rileyjmurray: RandBLAS defines an ``UNUSED(*)`` macro that you can invoke as ``UNUSED(layout)`` in the function body, and leave a note in the source for why you aren't using it. I'd prefer that to this ``[[maybe_unused]]`` syntax.

mmelnich: Fixed.

PR #138, opened 2026-05-12
FunNystrom++ V2

mmelnich:
WIP


rileyjmurray (changes requested):
I recommend you have Claude pull all changes from the benchmark/ files on this branch and move them to a new PR. Then you can reduce the scope of this PR to actually be about FunNystrom++.

PR #140, opened 2026-05-13
Bindings prep: RandLAPACK::Error, install.sh discovery, BQRRP validation

mmelnich:
Four coordinated changes preparing RandLAPACK for the bindings repo:

install.sh dependency discovery via BLASPP_INSTALL_DIR / LAPACKPP_INSTALL_DIR / RANDOM123_INSTALL_DIR env vars; skip clones and reuse existing installs when set. Also fixes a CUDA flag conflict when external blaspp depends on CUDAToolkit.

New RandLAPACK/rl_exceptions.hh defines RandLAPACK::Error plus a single `randlapack_require(cond) << "msg " << x;` macro backed by a RAII StreamThrower whose destructor throws at end of full expression. Registered in the umbrella header and CMakeLists.txt. The macro uses `for(; !(cond); )` rather than `if(cond);else` to short circuit without triggering -Wdangling-else inside unbraced outer ifs.

211 call sites across 19 RandLAPACK source files migrate from randblas_require to randlapack_require with hand readable messages that include the actual parameter values, e.g. `randlapack_require(b_sz >= 1) << "BQRRP::call: b_sz=" << b_sz << " must be positive"`. Polarity flips from "error if BAD" to "require GOOD"; (long long) casts drop since operator<< handles int64_t directly.

Driver entry point validation in every driver's call() and constructor, closing the segfault paths that bindings would expose: BQRRP (8 checks), RSVD (5), CQRRT (7), CQRRPT (8), REVD2 (6 across 2 overloads), ABRIK (9 across 3 overloads), krill_full_rpchol (6 new + 1 retained mu_size).

### DIFF
--- a/RandLAPACK/comps/rl_determiter.hh
+++ b/RandLAPACK/comps/rl_determiter.hh
@@ -384,7 +384,7 @@ void pcg(
     int64_t ns = n*s;
     int64_t ss = s*s;
     bool treat_as_separable = G.num_ops > 1;
-    if (treat_as_separable) randlapack_error_if_msg(s != G.num_ops, "with separable treatment, s=%lld must equal G.num_ops=%lld", (long long)s, (long long)G.num_ops);
+    if (treat_as_separable) randlapack_require(s == G.num_ops) << "with separable treatment, s=" << s << " must equal G.num_ops=" << G.num_ops;
 
     // All workspace gets zero-initialized; this is only
     // overridden for "R".
--- a/RandLAPACK/comps/rl_determiter.hh
+++ b/RandLAPACK/comps/rl_determiter.hh
@@ -1,5 +1,6 @@
 #pragma once
 
+#include "rl_exceptions.hh"
 #include "rl_blaspp.hh"
 
 #include <iostream>
@@ -383,7 +384,7 @@ void pcg(
     int64_t ns = n*s;
     int64_t ss = s*s;
     bool treat_as_separable = G.num_ops > 1;
-    if (treat_as_separable) randblas_require(s == G.num_ops);
+    if (treat_as_separable) randlapack_error_if_msg(s != G.num_ops, "with separable treatment, s=%lld must equal G.num_ops=%lld", (long long)s, (long long)G.num_ops);
 
     // All workspace gets zero-initialized; this is only
     // overridden for "R".
--- a/RandLAPACK/rl_exceptions.hh
+++ b/RandLAPACK/rl_exceptions.hh
@@ -1,0 +1,111 @@
+#pragma once
+
+#include <exception>
+#include <cstdarg>
+#include <cstdio>
+#include <string>
+
+// Most code in this file is structurally adapted from RandBLAS/exceptions.hh
+// (which itself was adapted from BLAS++). The macros and helpers here throw a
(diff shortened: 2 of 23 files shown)


mmelnich:
@rileyjmurray let me know what you think about this.

rileyjmurray:
I have a strict preference for randlapack_require as opposed to randlapack_error_if. When we read code it's much easier to incorporate statements of the form

> X is true.

than statements of the form

> if not X, then error.
> // ... only reachable if no error.

into our model model of what the code is doing.

RandLAPACK/comps/rl_determiter.hh
```diff
@@ -383,7 +384,7 @@ void pcg(
     int64_t ns = n*s;
     int64_t ss = s*s;
     bool treat_as_separable = G.num_ops > 1;
-    if (treat_as_separable) randblas_require(s == G.num_ops);
+    if (treat_as_separable) randlapack_error_if_msg(s != G.num_ops, "with separable treatment, s=%lld must equal G.num_ops=%lld", (long long)s, (long long)G.num_ops);
```
rileyjmurray: If you're going to have string formatting you should use C++ strings. I suggest you update your macro to handle stringstreams.

RandLAPACK/rl_exceptions.hh
```diff
@@ -0,0 +1,111 @@
...
+    (void) condstr;
+    if (cond) {
+        char buf[RandLAPACK_ERROR_MESSAGE_SIZE];
+        va_list va;
+        va_start(va, format);
+        vsnprintf(buf, sizeof(buf), format, va);
+        va_end(va);
+        throw Error(buf, func);
+    }
+}
```
rileyjmurray: Can we just use std::string instead of char arrays?

RandLAPACK/rl_exceptions.hh
rileyjmurray, line 1: Before acquiring technical debt from another project we should figure out if this approach (involving `char` arrays) is preferable to a std::string approach.

rileyjmurray (commented):
Sorry to be a pain, but I don't like the switch to ``randlapack_error_if(...)`` statements over ``randblas_require(...)`` statements. (Certainly fine to use a randlapack-defined `randlapack_require` statement over a `randblas_require` statement.)

RandLAPACK/rl_exceptions.hh
mmelnich: Fixed

RandLAPACK/comps/rl_determiter.hh
mmelnich: Fixed

RandLAPACK/rl_exceptions.hh
mmelnich: It is not, that was just a matter of c vs c++. Fixed.

mmelnich:
@rileyjmurray lmk what you think now.

rileyjmurray (approved):
:fire:

rileyjmurray:
The mechanism you're using for randlapack_require is very cool. (randblas_require statements are not nearly as helpful when they throw.) Maybe open a GitHub issue on the RandBLAS repo for RandBLAS to switch to this style.

PR #134, opened 2026-05-03
fix include paths

rileyjmurray:
Files affected by this PR only built in situations where people cloned the RandLAPACK repo as the name "RandLAPACK". It wouldn't work if someone (like me) cloned as
```
git clone <randlapack's URL> repo-randlapack
```
which pulls down the URL into a folder named "repo-randlapack".

This PR's changes make it so tests can build no matter the name of the folder containing someone's local copy of the repo.

Issue #74, opened 2024-09-07
C++ style memory allocation.


TeachRaccooon:
At the moment, I use `calloc` all throughout the repo.
Makes sense to change that.
Riley's comment:

_If you do new T[length_of_buffer]{} (not the curly braces), then it will zero-initialize.
An advantage to new is that it will work if T is std::complex<float> or std::complex<double> (defacto-standard datatypes for complex floats in C++). I don't know if calloc will work in that situation.
_That said, calloc apparently works with realloc, while new does not.
I says we open a GitHub issue about this so we establish a policy for new contributions.__

We also need to consider getting rid of the internal allocations.
The discussion on all that is present in PR#71

TeachRaccooon:
Fixed in PR #94

PR #157, opened 2026-08-12
TEMPORARY: quarantine the macOS Accelerate gesdd test (revert after Mark's BLAS++/LAPACK++ review)

mmelnich:
> [!WARNING]
> **This is a temporary change and is meant to be reverted.**
> It is one clean `git revert` of a single commit.

### Revert when these land

The quarantine exists only until Apple's new Accelerate interface is in use. Both are open and awaiting review:

| PR | Status |
|---|---|
| [icl-utk-edu/blaspp#134](https://github.com/icl-utk-edu/blaspp/pull/134) — New Apple Accelerate support (#74, rebased and completed) | open |
| [icl-utk-edu/lapackpp#88](https://github.com/icl-utk-edu/lapackpp/pull/88) — Support Apple's new Accelerate interface (closes #43) | open |

**Order matters:** lapackpp#88 depends on the `defines.h` from blaspp#134, so #134 merges first. blaspp#134 is a rebased and completed continuation of the stalled [blaspp#74](https://github.com/icl-utk-edu/blaspp/pull/74).

Once both are merged, [RandLAPACK#155](https://github.com/BallisticLA/RandLAPACK/pull/155) (retire legacy-Accelerate accommodations) becomes mergeable and `TestQB.Polynomial_Decay_general1` should pass on its own — at which point the warning added here fires and tells you to delete this quarantine.

## Why

`TestQB.Polynomial_Decay_general1` fails on Apple Silicon because Apple's old Accelerate LAPACK has a broken divide-and-conquer `gesdd`. It has been left failing deliberately, as a canary for the planned migration.

The side effect is that **`core-macos` has been permanently red** — `build` and `build-asan`, on `main` and on every branch. A lane that is always red carries no information: you cannot distinguish a new macOS regression from the known one, so the job gets ignored and the canary signals to nobody.

## What this does

Quarantines the test without discarding the signal:

- The main `ctest` run excludes it, so the job reflects real macOS health.
- The test is then **still run**, separately, and cannot fail the job.
- Every macOS run prints a warning that the suppression is active and names the condition for removing it.
- If the test ever **passes**, the job prints a louder warning saying to delete the quarantine.

That last point matters: the migration signal is preserved and actually improved. Previously it was "a red job turns green", which nobody was watching because the job was always red. Now it is an explicit annotation telling you what to delete.

## Scope

Deliberately surgical — `core-macos.yaml` (both jobs) and the corresponding `docs/CI.md` entry, nothing else, so the revert stays clean. The dead `TestABRIK.ABRIK_catch_instability` exclusion in these same lines is left alone for that reason.

### DIFF
--- a/.github/workflows/core-macos.yaml
+++ b/.github/workflows/core-macos.yaml
@@ -86,7 +86,27 @@ jobs:
               `pwd`/../RandLAPACK
           make -j$(sysctl -n hw.ncpu)
           make -j$(sysctl -n hw.ncpu) install
-          ctest --exclude-regex "^TestABRIK\.ABRIK_catch_instability" --output-on-failure
+          #####################################################################
+          ## TEMPORARY -- REVERT THIS ENTIRE BLOCK                            ##
+          ##                                                                  ##
+          ## TestQB.Polynomial_Decay_general1 fails on Apple Silicon because  ##
+          ## Apple's OLD Accelerate LAPACK has a broken divide-and-conquer    ##
+          ## gesdd. It is quarantined here ONLY to stop core-macos being      ##
+          ## permanently red, which trained everyone to ignore this job.      ##
+          ##                                                                  ##
+          ## REVERT AS SOON AS Mark has reviewed the BLAS++ / LAPACK++ work   ##
+          ## that migrates to the new Accelerate interface. The test is still ##
+          ## RUN below, so its result is not lost: if it starts PASSING, the  ##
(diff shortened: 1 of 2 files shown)


mmelnich:
Tracked for revert in #159 — the revert is to happen in #155, once [blaspp#134](https://github.com/icl-utk-edu/blaspp/pull/134) and then [lapackpp#88](https://github.com/icl-utk-edu/lapackpp/pull/88) are approved and merged.

PR #162, opened 2026-08-13
Installer overhaul: pins, provenance, a run-not-just-link BLAS check, backend selection, and docs

mmelnich:
## Problem

RandLAPACK's two installers had drifted apart. `install/install.ps1` was rebuilt in #156 and pins every dependency, checksums downloads, records provenance and link-and-run validates the BLAS before building anything. `install/install.sh` did none of that, and the documentation described a script that no longer existed in several respects.

Eleven defects, all verified against `main` before touching anything:

| Where | Defect |
|---|---|
| `install.sh:336,339,342` | three `git clone` with **no ref** — blaspp, lapackpp, random123 all tracked upstream default branches |
| — | **no BLAS validation at all** |
| — | `/opt/homebrew` hardcoded, **8 places** — hard-fails on Intel macOS and custom `HOMEBREW_PREFIX` |
| `install.sh:331` | `mv "$REPO_DIR"` relocated the user's own clone, breaking git worktrees |
| `install.sh:211` | `: > "$LOG"` truncated the log, destroying the previous run's output |
| `install.sh:247` | libdir search omitted `lib/aarch64-linux-gnu` |
| — | no `--prefix`; extras and benchmarks unconditional with no way to skip |
| `install.ps1` | **zero** references to `RANDNLA_PROJECT_DIR`, no `--modify-rc` equivalent |
| `INSTALL_SCRIPT.md` | documented the pre-backend flag list, and still told people the script would move their clone |

### Integer width is read back, not assumed

BLAS++ probes `int32` before `int64`, and `blas_int` only filters which *library names* to try. For MKL that is a real choice — `mkl_intel_lp64` and `mkl_intel_ilp64` are different libraries. For OpenBLAS there is only `-lopenblas`, so **a successful `blas_int=int64` configure proves nothing**: an LP64 build passes the `int32` probe and is accepted. The resolved width therefore comes from BLAS++'s generated `blas/defines.h` after the build. Trusting the request would stamp LP64 installs as ILP64 and have every later run reuse them believing otherwise.

### Verification runs the program, and that is the point

The conftest compiles, links and *runs* against the finished install — through BLAS++ and LAPACK++ rather than raw `dgemm_`, since that is how RandLAPACK reaches them — checking a `gemm` result and a `gesdd` factorization numerically. `gesdd` is chosen because it is the routine Apple's legacy Accelerate computes incorrectly.

Running it rather than only linking matters for a reason specific to integer width. BLAS++ and LAPACK++ *do* guard the `int64_t` downcast — `to_blas_int` (`blaspp/src/blas_internal.hh:16-22`) throws rather than truncates — but that guard keys off `sizeof(blas_int)` **as declared by the header**. If the headers say 64-bit while the library actually loaded is LP64, the guard compiles out and 64-bit values reach routines reading 32 bits. That surfaces not as wrong numbers but as nonsense control values: a misread workspace query becomes an absurd `lwork`, and the run dies in allocation or never finishes. Nothing catches it by inspection.

## A cross-project collision, found while testing

Dependency install directories now carry backend and GPU in their names (`blaspp-mkl-cpu-install`). That is not cosmetic.

RandBLAS's installer uses the same `RandNLA-project` layout and named its BLAS++ install **identically** (`blaspp-<backend>-install`), with a different stamp filename. With a shared `RANDNLA_PROJECT_DIR`, this script would rebuild over RandBLAS's BLAS++ — their stamp never matches ours — and **RandBLAS's next run would reuse an artifact we had replaced underneath it, while its own stamp still described the original.** Distinct names make it impossible; sharing is instead explicit via `BLASPP_INSTALL_DIR`, verified by the conftest.

## Everything else

- **macOS keeps Homebrew OpenBLAS as the default.** Apple's legacy Accelerate has a broken divide-and-conquer `gesdd`, and RandLAPACK calls `gesdd` in `rl_rsvd.hh:146`, `rl_abrik.hh:692`, `rl_revd2.hh:208`, `rl_preconditioners.hh:355` and `rl_util.hh:413`. That is what #157's quarantine is about. `--blas=accelerate` warns, citing #159. The hardcoded `/opt/homebrew` is fixed regardless.
- **Extras and benchmarks stay built by default** (`--no-extras` / `--no-benchmarks`), unlike RandBLAS's examples — they need nothing this script has not already built.
- **The clone is never moved**; a symlink gives the same layout and the path CI invokes still resolves.
- **RandBLAS is untouched** — still a pinned submodule, still authoritative.
- **`install.ps1`** now honours `RANDNLA_PROJECT_DIR` with the same precedence as `install.sh`, gains `-ModifyEnvironment` (User-scope, the only thing that survives a new shell) and `-Prefix`, and its path-length warning now checks the resolved path rather than only an explicitly passed one.
- **Progress rendering** in three tiers, determinate from the build tool's own output, with tier 0 byte-identical to the plain step list — plus a CI assertion that redirected output carries no escape sequence.
- **`INSTALL_SCRIPT.md`** rewritten: current flags, corrected layout diagram, and a new section 6 with a tested-configuration table drawn from the CI lanes, the per-backend integer width, whether LP64 actually limits RandLAPACK, and why macOS is on OpenBLAS.

## Verification

Local Linux (gcc 15.2, oneAPI MKL, NVIDIA GPU present):

| Scenario | Result |
|---|---|
| Fresh MKL build, `--no-gpu` | 9/9 steps, ILP64 confirmed from `blas/defines.h` |
| Re-run in place | all six dependency steps reused, 1s rebuild |
| Full default run (extras + benchmarks) | 13/13 steps |
| `--no-gpu` → `--gpu` | BLAS++ provenance invalidated, rebuilt rather than reused |
| Dependency discovery | all three reused, 3/3 steps |
| Run from a git worktree | clone left in place, worktree still functional |
| Output redirected | no ANSI escapes, no carriage returns |
| **Full test suite** | **313/313 pass** |
| RandBLAS then RandLAPACK, shared `RANDNLA_PROJECT_DIR` | both dependency sets intact |
| RandLAPACK reusing RandBLAS's BLAS++ | reused, only LAPACK++ built, 5/5 steps |

Windows 11, Windows PowerShell 5.1, VS 2022 Build Tools: all three `-ProjectDir` precedence cases resolve as intended, and the User-scope environment write round-trips.

## RandBLAS submodule bump

Also bumps the vendored RandBLAS from `04f2018` to `952251c` (RandBLAS main), and `RandLAPACK_RandBLAS_PIN` with it.

The three commits in that range are **#185** (vcpkg manifest fix), **#186** (CMake hygiene) and **#187** (installer scripts). None of them touch `RandBLAS/` headers — the library code is byte-identical, and everything in the range is build machinery. Two pieces of it matter here:

- RandBLAS's CMake floor moved 3.12 → 3.21. RandLAPACK already declares 3.21, so nothing to change; the submodule now simply enforces what RandLAPACK already required.
- RandBLAS exports `randblas_stage_runtime_dlls()` from its installed package and prints a configuration summary. As a subproject under RandLAPACK, which configures it with `BUILD_TESTS=OFF`, that summary reports tests as deliberately skipped rather than warning.

`CMakeLists.txt` enforces that the pin variable and the submodule move together, and it means it: the cross-check compares the variable against `git ls-tree HEAD RandBLAS`, so a staged-but-uncommitted bump fails configure until both land in one commit. Worth knowing if you ever bump this iteratively — it is not satisfiable before committing.

Verified: clean build against the bumped submodule, and the full suite re-run.

## Notes for reviewers

- **Consolidated from #162 + #163 + #164** at Max's request, matching how the RandBLAS installer set was combined. Those two show as *merged* rather than closed because this branch was fast-forwarded to contain them — they merged into their stacked base, not into `main`. The tree is byte-identical to the state tested and green as three separate PRs.
- Stacked on #161 (CI fixes), which stays separate — it is independent and useful on its own.
- **`-ModifyEnvironment` is not verified end to end.** It runs at the end of a successful install, and I could not reach it: the clone sits on a `\\wsl.localhost` path and Windows `git` refuses it with `dubious ownership`. Precedence and the User-scope write are each verified directly, but not together in one completed run.
- Two upstream gaps are filed rather than worked around silently: BLAS++ never looks for an ILP64 OpenBLAS (#166), and implements only Apple's legacy Accelerate interface (#165).

### DIFF
--- a/.github/workflows/install-script.yaml
+++ b/.github/workflows/install-script.yaml
@@ -56,26 +56,47 @@ jobs:
           path: deps-install
           key: installer-deps-${{ runner.os }}-v1
 
+      # Globbed rather than spelled out: dependency install directories carry
+      # the backend and GPU configuration in their names (blaspp-openblas-cpu-install),
+      # so hardcoding them here would couple CI to that naming and break the
+      # moment it changes.
       - name: seed discovery variables from cache
         if: steps.deps-cache.outputs.cache-hit == 'true'
         run: |
-          echo "BLASPP_INSTALL_DIR=$GITHUB_WORKSPACE/deps-install/blaspp-install" >> "$GITHUB_ENV"
-          echo "LAPACKPP_INSTALL_DIR=$GITHUB_WORKSPACE/deps-install/lapackpp-install" >> "$GITHUB_ENV"
+          set -euo pipefail
+          echo "BLASPP_INSTALL_DIR=$(echo "$GITHUB_WORKSPACE"/deps-install/blaspp-*-install)" >> "$GITHUB_ENV"
+          echo "LAPACKPP_INSTALL_DIR=$(echo "$GITHUB_WORKSPACE"/deps-install/lapackpp-*-install)" >> "$GITHUB_ENV"
           echo "RANDOM123_INSTALL_DIR=$GITHUB_WORKSPACE/deps-install/random123" >> "$GITHUB_ENV"
 
       - name: keep a pristine clone for the discovery test
         run: cp -a RandLAPACK RandLAPACK-discovery
 
       - name: run the installer (non-interactive)
-        run: bash RandLAPACK/install.sh --yes --no-gpu
+        run: |
+          set -euo pipefail
+          bash RandLAPACK/install.sh --yes --no-gpu 2>&1 | tee installer.out
+
+      # Redirected output must carry no ANSI escapes and no carriage returns.
+      # The installer draws a progress bar on a terminal; without this check,
+      # that bar could start filling install.log and every CI transcript with
+      # control characters and nobody would notice until a log was unreadable.
+      - name: piped output is free of terminal control sequences
+        run: |
+          if LC_ALL=C grep -qP '\x1b\[|\r' installer.out; then
+            echo "Found terminal control sequences in non-TTY output:"
+            LC_ALL=C grep -nP '\x1b\[|\r' installer.out | head -20
+            exit 1
+          fi
+          echo "OK: no escape sequences in redirected output."
 
       - name: populate the dependency cache
         if: steps.deps-cache.outputs.cache-hit != 'true'
         run: |
+          set -euo pipefail
           mkdir -p deps-install
-          cp -a RandNLA-project/install/blaspp-install   deps-install/
-          cp -a RandNLA-project/install/lapackpp-install deps-install/
-          cp -a RandNLA-project/install/random123        deps-install/
+          cp -a RandNLA-project/install/blaspp-*-install   deps-install/
+          cp -a RandNLA-project/install/lapackpp-*-install deps-install/
+          cp -a RandNLA-project/install/random123          deps-install/
 
       - name: test the installed library
         run: |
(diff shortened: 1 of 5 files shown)


mmelnich:
Test suite result, as promised: **313/313 pass, 0 failed** on the build this installer produced (Linux, gcc 15.2, oneAPI MKL, ILP64, `--no-gpu`). All CI checks green as well.

For the record, since the run took a while and it is the slow tests that matter here: `TestCQRRT.CQRRT_large_full_rank` alone is ~1080s and several ABRIK/REVD2 cases are 60-70s each, so a local full-suite run against an installer build is around 40 minutes wall clock, not something to expect in a quick loop.

PR #163, opened 2026-08-13
install.ps1: honour RANDNLA_PROJECT_DIR, and add -Prefix and -ModifyEnvironment

mmelnich:
## Problem

`install.ps1` read `RANDNLA_PROJECT_DIR` **nowhere**. That variable is the shared convention between this installer, `install.sh` and RandBLAS's installer — it is what lets one machine keep a single `RandNLA-project` tree instead of scattering several. On Windows it simply did not exist, so `-ProjectDir` was the only way to say where anything went, and there was no equivalent of `install.sh`'s `--modify-rc` to make a choice persist.

There was also no `-Prefix`, so RandLAPACK could only ever install to `<ProjectDir>\install\RandLAPACK-install`.

## What this PR does

1. **`RANDNLA_PROJECT_DIR` is honoured**, with precedence identical to `install.sh`: the flag, then the environment variable, then a sibling of the clone.
2. **`-ModifyEnvironment`** persists it via `[Environment]::SetEnvironmentVariable(..., "User")` — the Windows equivalent of appending to a shell profile, and the only mechanism that survives opening a new shell (setting `$env:` alone lasts for the current process). Opt-in, matching `install.sh`: the default touches nothing and prints the `setx` line.
3. **`-Prefix`** installs RandLAPACK itself elsewhere — a module tree, or any prefix a site wants to own. Dependencies stay in the project directory.
4. **Moved the path-length warning to after resolution.** It was guarded on `$ProjectDir -ne ""`, so it only fired for an explicitly passed `-ProjectDir` — never for the default, which is derived from wherever the clone happens to sit and is therefore the more likely one to be long.

## Verification

On Windows 11, **Windows PowerShell 5.1**, VS 2022 Build Tools:

| Case | Resolved to |
|---|---|
| no env var, no flag | `...\worktrees\RandNLA-project` (sibling default) |
| `$env:RANDNLA_PROJECT_DIR = C:\rl-from-env` | `C:\rl-from-env` |
| `-ProjectDir C:\rl-from-flag` with the env var still set | `C:\rl-from-flag` |
| `SetEnvironmentVariable(..., "User")` round-trip | written and read back correctly |

Worth noting incidentally: those runs exercised the #156 machinery on the way through, and oneMKL discovery, the BLAS/LAPACK link check, and the space-free import-library staging (for `C:\Program Files (x86)\Intel\oneAPI\...`) were all observed still working.

## Notes for reviewers

- **What I could not test end to end, and why.** `-ModifyEnvironment` runs at the *end* of a successful install, and I could not reach it from this setup: the clone lives on a `\\wsl.localhost` path, and Windows `git` refuses it with `fatal: detected dubious ownership`, which stops the dependency step. So the precedence logic and the User-scope write are each verified directly, but not the two together in one completed run. Worth someone doing on a Windows-native clone before merge, or I can if you'd rather.
- Stacked on #162, which is stacked on #161. Retarget as those merge.
- No behaviour changes to backend selection, DLL staging or the toolchain guard — everything from #156 is untouched.

### DIFF
--- a/install/install.ps1
+++ b/install/install.ps1
@@ -10,8 +10,13 @@
 # Windows yet.
 #
 # Options:
-#   -ProjectDir <path>  Where dependencies/builds/installs go
-#                       (default: ..\RandNLA-project next to the clone).
+#   -ProjectDir <path>  Where dependencies/builds/installs go. Defaults to
+#                       $env:RANDNLA_PROJECT_DIR when set, otherwise
+#                       ..\RandNLA-project next to the clone.
+#   -Prefix <path>      Install RandLAPACK itself here instead of
+#                       <ProjectDir>\install\RandLAPACK-install.
+#   -ModifyEnvironment  Persist RANDNLA_PROJECT_DIR for your user account. The
+#                       default touches nothing and prints the setx command.
 #   -Backend <name>     mkl (default) | openblas | custom. See setup.ps1.
 #   -MklRoot <path>     Use this oneMKL install instead of discovery.
 #   -NoDownload         Fail rather than download a backend that was not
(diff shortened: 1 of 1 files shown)


mmelnich:
For the record: this shows as merged rather than closed because #162's branch was fast-forwarded to contain it, so its commits are genuinely in that branch — it merged into its stacked base, not into `main`. Nothing here reached `main` independently.

Max asked for #162, #163 and #164 to land as one PR, matching how the RandBLAS installer set was consolidated. All of this work is now reviewable in **#162**.

PR #164, opened 2026-08-13
INSTALL_SCRIPT.md: document the new options, and add tested configurations

mmelnich:
## Problem

`INSTALL_SCRIPT.md` described a script that no longer exists in several respects, and omitted the things people most need to know.

- The **flag list predated backend selection entirely** — no `--blas`, `--blas-int`, `--prefix`, or the extras/benchmark opt-outs.
- The **directory diagram said the script would move your clone** into `lib/`. It no longer does (that broke git worktrees), and `lib/RandLAPACK` is a symlink.
- **"What the Script Does"** described cloning dependencies from their default branches — they are now pinned — and moving the RandLAPACK directory.
- There was **no statement of which configurations are tested**. You had to read the workflow files.
- There was **nothing on integer width**, which is the setting most likely to be wrong without anyone noticing.

## What this PR does

Documentation only; no code changes.

1. **Flag list brought current**, with extras and benchmarks stated as built by default with opt-outs — deliberately unlike RandBLAS's `examples/`, which are opt-in because they pull in dependencies RandBLAS itself does not need.
2. **Directory diagram corrected** to show the symlink and the `install/` tree, whose dependency directories carry backend and GPU configuration in their names.
3. **"What the Script Does" rewritten** around the pinned fetches, the provenance stamps that make reuse configuration-aware, and the verification step — including why that step *runs* a program rather than only linking one.
4. **New section 6**, which is the substantive addition:
   - **6.1** a tested-configuration table, every row a CI lane, so it says what is exercised rather than what ought to work. Compiler floor (gcc ≥ 13, for RandBLAS's C++20 concepts) stated, and CUDA noted as locally-tested-only since there is no GPU runner.
   - **6.2** per-backend integer width, with the OpenBLAS case explained: BLAS++ probes `int32` before `int64` and `blas_int` only filters library *names*, so with one candidate name an LP64 OpenBLAS passes the probe and is accepted — a successful `blas_int=int64` configure proves nothing. Includes the `--blas=custom` recipe for an ILP64 OpenBLAS, which exists but which BLAS++ never looks for.
   - **6.3** whether LP64 actually limits RandLAPACK. Rarely, and loudly: `to_blas_int`/`to_lapack_int` throw rather than truncate, and the guard is on individual dimensions rather than element counts, so a 100,000 × 100,000 matrix is fine. The genuinely *silent* case — headers built for one width, library loaded with the other — is called out separately, since that is the one the verification step exists to catch.
   - **6.4** why macOS defaults to Homebrew OpenBLAS rather than Accelerate: Apple's legacy Accelerate has a broken divide-and-conquer `gesdd`, and RandLAPACK calls `gesdd` in five library files, so most SVD-based drivers would be affected. Cross-referenced to #159.

## Notes for reviewers

- **This overlaps #153**, which moves this file to `docs/INSTALL_SCRIPT.md` and also edits the root `install.sh` wrapper. I have deliberately updated the copy that exists on `main` so this PR stands alone; #153 will need to absorb it on rebase. Flagging rather than letting the merge discover it. #155 also edits root `INSTALL.md`, so three branches currently touch install docs.
- The table in 6.1 is only worth having if it stays true. If a CI lane is added or dropped, that row should move with it.
- Stacked on #163 → #162 → #161. Retarget as those merge.

### DIFF
--- a/INSTALL_SCRIPT.md
+++ b/INSTALL_SCRIPT.md
@@ -90,15 +90,26 @@ The install script expects a specific directory structure:
 
 ```
 ~/RandNLA/
-├── RandLAPACK/          # Clone RandLAPACK here (script will move it)
-└── RandNLA-project/     # Created automatically by script
+├── RandLAPACK/               # Your clone. The script does NOT move it.
+└── RandNLA-project/          # Created automatically
     ├── lib/
-    │   ├── blaspp/      # Built by script
-    │   ├── lapackpp/    # Built by script
-    │   ├── random123/   # Built by script
-    │   └── RandLAPACK/  # Moved here by script
-    └── build/           # Build artifacts
+    │   ├── blaspp/           # Source, fetched at a pinned commit
+    │   ├── lapackpp/         # Source, fetched at a pinned commit
+    │   └── RandLAPACK -> ../../RandLAPACK   # Symlink to your clone
+    ├── install/
+    │   ├── blaspp-<backend>-<cpu|cuda>-install
+    │   ├── lapackpp-<backend>-<cpu|cuda>-install
+    │   ├── random123/
+    │   └── RandLAPACK-install
+    └── build/                # One build directory per project above
 ```
+
+Two things worth noting. **Your clone stays where you put it** — earlier
+versions of this script relocated it into `lib/`, which broke git worktrees;
+`lib/RandLAPACK` is now a symlink. And the dependency install directories carry
+the backend and GPU configuration in their names, so an ILP64 MKL build and an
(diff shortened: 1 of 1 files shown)


mmelnich:
For the record: this shows as merged rather than closed because #162's branch was fast-forwarded to contain it, so its commits are genuinely in that branch — it merged into its stacked base, not into `main`. Nothing here reached `main` independently.

Max asked for #162, #163 and #164 to land as one PR, matching how the RandBLAS installer set was consolidated. All of this work is now reviewable in **#162**.

PR #175, opened 2026-09-20
Fix numerical-rank counting and scratch-buffer cleanup

rileyjmurray:
`util::rank_check` undercounts rank-deficient matrices by one and returns -1 for a zero matrix. Its early return also skips both scratch-buffer deletions. This PR keeps the threshold unchanged and returns the number of singular values above it after releasing the buffers.

I've extracted Maksim Melnichenko's fix and regression tests from #130 so they can merge before the block Krylov work.

Tests: all 18 utility tests pass with Spack LLVM 19 and Accelerate LAPACK 3.12. Before the fix, the new rank-7 and zero-matrix tests returned 6 and -1, respectively; the full-column-rank case already passed.


RandLAPACK/misc/rl_util.hh
rileyjmurray, line 1: Comments about what code _previously_ did rarely belong in source code. Remove.

rileyjmurray:
This can merge once tests pass.

PR #178, opened 2026-09-20
Add validated dense text and binary matrix readers

rileyjmurray:
This extracts the dense text and binary readers from Maksim Melnichenko's #130 for use by matrix generators and other callers. Both use a dimension query followed by a read into caller-provided storage, with a choice of row-major or column-major layout and a leading dimension. The binary format is two native-endian int64 dimensions followed by row-major doubles.

The readers report malformed input and invalid output dimensions or strides through `RandLAPACK::Error`. They preserve padding and check the shape supplied for the second call, preventing changed binary dimensions from overrunning a buffer allocated after the query. The existing `process_input_mat` entry point remains as a column-major forwarding wrapper, including its query-flag protocol.

Tests: all 336 RandLAPACK tests pass with Spack LLVM 19 and Accelerate LAPACK 3.12, including 19 matrix-I/O cases covering both layouts, padded storage, malformed input, invalid or overflowing strides, and the generator and legacy query paths.


RandLAPACK/testing/rl_matrix_io.hh
```diff
@@ -0,0 +1,165 @@
...
+#include <limits>
+#include <stdexcept>
+#include <string>
+#include <vector>
+
+namespace RandLAPACK::gen {
+
+namespace matrix_io_detail {
+
+[[noreturn]] inline void fail(const char* filename, const char* reason) {
```
rileyjmurray, line 21: Why the `[[noreturn]]`` annotation?

RandLAPACK/testing/rl_matrix_io.hh
```diff
@@ -0,0 +1,165 @@
...
+    throw std::runtime_error(std::string(reason) + ": " + (filename ? filename : "(null)"));
```
rileyjmurray: I think RandLAPACK has its own exception class. Push back if I'm wrong.

RandLAPACK/testing/rl_matrix_io.hh
```diff
@@ -0,0 +1,165 @@
...
+        return;
+    }
+
+    file.seekg(payload_start);
+    std::vector<double> values(count);
+    if (!file.read(reinterpret_cast<char*>(values.data()), bytes))
+        fail(filename, "Failed to read matrix payload");
+    for (int64_t row = 0; row < m; ++row)
+        for (int64_t col = 0; col < n; ++col)
+            A[row + m * col] = static_cast<T>(values[row * n + col]);
```
rileyjmurray: Include an empty return statement at the end of the function. People should be able to set a breakpoint when the function's work is done but we're still in its stack frame.

RandLAPACK/testing/rl_matrix_io.hh
```diff
@@ -0,0 +1,165 @@
...
+/// Read a nonempty dense matrix from a native-endian binary file.
+/// The file contains two int64_t dimensions (m, n), followed by exactly m*n
+/// IEEE-754 64-bit doubles in row-major order. Files must use the reader's byte
+/// order; this format has no byte-order marker or automatic byte swapping.
+///
+/// With query=true, validate dimensions and payload size, then replace m and n;
+/// A may be nullptr. Otherwise, m and n must match the header and A must point
+/// to at least m*n elements. The dimensions are unchanged during a data read.
+/// Values are converted to T and stored column-major: A[row + m*col].
+/// Invalid dimensions, size mismatches, and read errors throw std::runtime_error.
```
rileyjmurray: Strange to require writing in row-major order and then read back into column-major storage. This function should accept a layout parameter for A as well as a stride parameter `lda`. It's fine to assume row-major for how the data is stored on disk.

RandLAPACK/testing/rl_matrix_io.hh
rileyjmurray: The annotation records that `fail` never returns normally: it always throws. It lets the compiler account for that in control-flow analysis; exceptions still unwind and can be caught. I kept it.

--- Astra

RandLAPACK/testing/rl_matrix_io.hh
rileyjmurray: `RandLAPACK::Error` is defined in `rl_exceptions.hh`. The reader now includes that header directly and uses the library's exception type; the error-path tests check it too.

--- Astra

RandLAPACK/testing/rl_matrix_io.hh
rileyjmurray: Added a final `return;` to both readers and the `process_input_mat` wrapper so each has an explicit end-of-work line for a breakpoint.

--- Astra

RandLAPACK/testing/rl_matrix_io.hh
rileyjmurray: Both readers now take `Layout` and `lda`. They support row-major or column-major destinations, preserve padding, and reject insufficient or overflowing strides before writing. Query calls ignore `lda` and accept a null output buffer. Binary payloads remain row-major on disk. Tests cover both layouts with compact and padded storage for `float` and `double`.

--- Astra

PR #179, opened 2026-09-20
Fix the adversarial CholeskyQR matrix generator

rileyjmurray:
The adversarial CholeskyQR generator returns an all-ones spectrum because its decay loop starts at its end. This PR uses `frac_spectrum_one` to set the leading block, followed by a drop to `sqrt(eps<T>)` and geometric decay to `1/cond`. The minimum condition number is `1/sqrt(eps<T>)` for the requested numerical type.

The generator fix and original spectrum tests come from Maksim Melnichenko's work in #130. This PR also validates parameters and corrects rectangular diagonal output. New overloads take the leading fraction explicitly; the old signatures retain the default fraction of 0.1. `mat_gen` supplies its existing `frac_spectrum_one` field. Diagonal output writes the spectrum directly into the destination, avoiding the temporary square matrix.

Tests: all 334 tests pass with Spack LLVM 19 and Accelerate LAPACK 3.12, including 23 generator tests. Coverage includes float and double bounds, rectangular output, random singular vectors, and legacy signatures. A CQRRPT regression constructs an exactly rank-deficient input with duplicate columns from the generated spectrum, checking pivoting, reconstruction of nonzero dependent columns, and orthogonality. Ten generator tests fail against the implementation preceding the precision fix because of its fixed threshold and spectral cliff.

--- Astra


RandLAPACK/testing/rl_gen.hh
```diff
@@ -358,26 +359,61 @@ void gen_oleg_adversarial_mat(
...
+///                               least two.
+/// @param[in] cond               Target condition number. Must be >= 1e8.
+///
+/// @return Vector of k singular values, non-increasing, with s[0] = 1 and
+///         s[k-1] = 1/cond.
 template <typename T>
-std::vector<T> gen_bad_cholqr_singvals(int64_t k, int64_t n, T cond) {
+std::vector<T> gen_bad_cholqr_singvals(int64_t k, T frac_spectrum_one, T cond) {
+    if (k < 3)
+        throw Error("k must allow one leading and two decaying values");
```
rileyjmurray: What's this `Error` class? Use something defined in RandLAPACK.

RandLAPACK/testing/rl_gen.hh
```diff
@@ -358,26 +359,61 @@ void gen_oleg_adversarial_mat(
...
+/// @param[in] k                  Number of singular values to generate.
+/// @param[in] frac_spectrum_one  Fraction of the spectrum held at 1.0, as in
+///                               gen_poly_singvals. Must leave a leading block of
+///                               at least one entry and a trailing block of at
+///                               least two.
+/// @param[in] cond               Target condition number. Must be >= 1e8.
+///
+/// @return Vector of k singular values, non-increasing, with s[0] = 1 and
+///         s[k-1] = 1/cond.
 template <typename T>
```
rileyjmurray, line 383: Hard-coding the min condition number to `1e8` while allowing general real types is inappropriate. We should set the lower bound at 1/sqrt(eps<T>) for the numerical type T.

RandLAPACK/testing/rl_gen.hh
```diff
@@ -386,24 +422,36 @@ void gen_bad_cholqr_mat(
...
 
     if (diagon) {
-        lapack::lacpy(MatrixType::General, k, k, S, k, A, k);
+        lapack::laset(MatrixType::General, m, n, T(0), T(0), A, m);
+        lapack::lacpy(MatrixType::General, k, k, S, k, A, m);
     } else {
         RandLAPACK::gen::gen_singvec(m, n, A, k, S, state);
     }
 
     delete[] S;
```
rileyjmurray: Do we really need `S` on all codepaths?

RandLAPACK/testing/rl_gen.hh
rileyjmurray: `Error` already resolves to `RandLAPACK::Error`, defined in [`rl_exceptions.hh`](https://github.com/BallisticLA/RandLAPACK/blob/291b9dcfd30161c60ed7d831a5e5aa12cd481578/RandLAPACK/rl_exceptions.hh#L37). In 291b9dc I qualified all three explicit throws as `RandLAPACK::Error` and expanded the tests for the invalid spectrum shapes.

--- Astra

RandLAPACK/testing/rl_gen.hh
rileyjmurray: Fixed in 291b9dc. The lower bound is now `1/sqrt(std::numeric_limits<T>::epsilon())`, and the spectral cliff moves to `sqrt(eps<T>)`. Interpolation uses `cond/min_cond` so rounding at the lower bound cannot make the trailing block increase. Float and double tests cover the bound, the value immediately below it, monotonicity, and the realized condition number. All 333 tests pass locally with the Spack toolchain, including 23 generator tests.

--- Astra

RandLAPACK/testing/rl_gen.hh
rileyjmurray: The diagonal path does not need `S`. In 291b9dc it zeros `A` and copies the singular values directly onto its diagonal. The temporary square matrix is now allocated only for random singular vectors, with `std::vector` managing its lifetime. Tests cover prefilled tall and wide rank-deficient diagonal output and the random-vector path.

--- Astra

rileyjmurray:
Added `CQRRPT_bad_cholqr_exact_rank_deficient` in aeb05c2. It uses the generator's spectrum `{1, 1, 2^-26, 2^-28}` in a 64-by-16 matrix with four dense columns interleaved among twelve zero columns. Walsh transforms preserve the spectrum in exact binary64 arithmetic; QR keeps the zero columns exactly zero, avoiding a rounding-sensitive rank decision. The test checks rank, pivots, reconstruction, and orthogonality over eight sketch seeds without prescribing pivot order. Locally, 1,024 seeds each under nearest, upward, downward, and toward-zero rounding all recovered rank four: worst relative Frobenius residual `9.41e-16`, worst `||Q'Q-I||_F/sqrt(4)` `4.61e-15`. These are measurements with Spack LLVM 19 and Accelerate on this machine. All 334 tests pass.

--- Astra

rileyjmurray:
Updated the CQRRPT regression in 2aaad81. Each active column now has an exact copy, and pivot checks accept either member of each pair. Helpers outside the fixture separate construction, pivot validation, reconstruction, and orthogonality; the test body is 27 lines. The revised test catches deliberately omitted trailing columns of R with relative residual 0.707, versus its tolerance of 1.82e-12. All 334 tests pass, and a local probe passed 1,024 seeds under each of four rounding modes with Spack LLVM 19 and Accelerate. The rank decision for duplicate columns now involves rounding, so these results provide local evidence rather than a guarantee across implementations.

--- Astra

PR #180, opened 2026-09-20
Support linear operators in randomized SVD

rileyjmurray:
RSVD can operate on sparse and implicit matrices without materializing a dense copy. This extracts Maksim Melnichenko's operator-based RS/RF/QB/RSVD path from #130, with a downdatable operator that applies accumulated QB deflation as low-rank corrections. The caller supplies the Frobenius norm used by QB's stopping test.

The existing dense calls and constructors retain their behavior. Operator calls require the concrete RS/RF/QB components and reject incompatible algorithm objects. The downdatable operator handles both storage layouts and sizes its scratch buffer from the actual number of right-hand sides. Failures in range finding or orthogonalization propagate through RSVD before attempting a small SVD. LAPACK failures also preserve the output pointers and release temporary buffers.

Tests: all 328 RandLAPACK tests pass with Spack LLVM 19 and Accelerate LAPACK 3.12. The eleven new cases cover float/double dense and sparse reconstruction, partial blocks, power-iteration parity, layout and transpose combinations, wide right-hand sides, capacity checks, failure handling, and input preservation. The five affected headers compile individually with `-Werror`.


RandLAPACK/linops/rl_downdatable_linop.hh
rileyjmurray, line 1: Rename the factors in downdatable linop from `(Q, BT)` to `(E, F)`. The current choice of `B` conflicts with the standard signature of `operator()` and `Q` wrongly suggests that the left factor needs to be column-orthonormal.

test/linops/test_downdatable_linop.cc
rileyjmurray, line 1: This is not even a little bit readable. It needs to be redone in its entirety.

test/drivers/test_rsvd_linop.cc
```diff
@@ -0,0 +1,161 @@
+#include "rl_rs.hh"
+#include "rl_rsvd.hh"
+#include "rl_sparse_linop.hh"
+#include <gtest/gtest.h>
+
+#include <cstdlib>
+#include <vector>
+
+namespace {
+using RNG = r123::Philox4x32;
```
rileyjmurray: Here and elsewhere, use `RNG = RandBLAS::DefaultRNG`.

RandLAPACK/drivers/rl_rsvd.hh
```diff
@@ -153,4 +175,62 @@ int RSVD<T, RNG>::call(
     return 0;
 }
 
+// -----------------------------------------------------------------------------
+// LinOp-templated RSVD: accepts any LinearOperator.
+// The base operator is never modified: deflation is handled implicitly
+// by DowndatableLinOp inside QB.
```
rileyjmurray, line 213: The behavior of this function in terms of exceptions vs return codes is unclear, and seems inconsistent.

rileyjmurray (commented):
This implementation has a bunch of free functions with comments saying things need to be kept in sync with class-based methods. The reason given has something to do with the `virtual` qualifier on the `call` functions. RandLAPACK is still at a development status where we can change the API at will. I'd like an explanation of the pros/cons of (and alternatives to) the current class-based API. Please prepare a detailed Beamer presentation on this point and upload to github as an attachment to a comment.

rileyjmurray:
The requested 28-slide Beamer presentation is attached:

- [Algorithm API design for linear operators (PDF)](https://github.com/user-attachments/files/32428444/randlapack-pr180-api-design.pdf)
- [Editable Beamer source and build instructions (ZIP)](https://github.com/user-attachments/files/32428445/randlapack-pr180-api-design-source.zip)

The presentation separates algorithm objects, strategy dispatch, and matrix representation. C++ forbids virtual member templates, but permits nonvirtual template overloads alongside virtual methods. The current dense-only strategy interfaces explain the casts and duplicated paths; neither classes nor QB deflation require duplicated RS/RF kernels.

It compares shared kernels with adapters, concrete strategy composition, erased operator interfaces, variants, and CRTP, with code sketches, tradeoffs, migration steps, and tests. My recommendation is concrete composition for the new C++ API unless open-ended runtime strategy replacement is a requirement; in that case, use a fixed erased-operator signature. The analysis links to the reviewed revision, and identifies proposed APIs as sketches.

The four code-review follow-ups are implemented in [951aa4a](https://github.com/BallisticLA/RandLAPACK/commit/951aa4a94817ff162c27cb004d21cb1e952fb155), including corrected RS/RF comments. Validation used the workspace's Spack tooling: 351/351 CTest cases passed, the final focused rerun passed 33/33, and all five affected headers compile independently with `-Werror`. The fixes received an independent code review; every presentation slide was rendered and visually checked.

--- Astra

RandLAPACK/linops/rl_downdatable_linop.hh
rileyjmurray: Renamed the stored factors, update parameters, buffer sizes, and formulas to `E` and `F` in [951aa4a](https://github.com/BallisticLA/RandLAPACK/commit/951aa4a94817ff162c27cb004d21cb1e952fb155). The class now explicitly documents `A - E F^T`, with no orthogonality requirement on either factor. The tests use nonorthogonal factors.

--- Astra

test/linops/test_downdatable_linop.cc
rileyjmurray: Rewrote the file in [951aa4a](https://github.com/BallisticLA/RandLAPACK/commit/951aa4a94817ff162c27cb004d21cb1e952fb155). The application checks are now 16 individually named parameterized cases covering both layouts, both transpose flags, and one or seven RHS columns. Small helpers separate storage, RHS construction, and the scalar reference calculation; the expected residual is a hand-derived matrix.

The cases retain successive updates and nonunit alpha/beta, and check padded storage, unchanged base/RHS buffers, and untouched output padding. Rank-capacity, factor-buffer, and scratch-buffer overflow each have their own test. All 19 downdatable-operator cases pass.

--- Astra

test/drivers/test_rsvd_linop.cc
rileyjmurray: Changed the RNG aliases to `using RNG = RandBLAS::DefaultRNG` in the operator RSVD tests and the affected dense RSVD and QB tests in [951aa4a](https://github.com/BallisticLA/RandLAPACK/commit/951aa4a94817ff162c27cb004d21cb1e952fb155).

--- Astra

RandLAPACK/drivers/rl_rsvd.hh
rileyjmurray: Made the operator overload's contract explicit in [951aa4a](https://github.com/BallisticLA/RandLAPACK/commit/951aa4a94817ff162c27cb004d21cb1e952fb155), following the policy we settled on:

- Invalid inputs or incompatible components throw `RandLAPACK::Error`.
- Numerical failures return named `RSVD::Status` values: `BlockOrthogonalityFailure` (4), `BasisOrthogonalityFailure` (5), `QBSubroutineFailure` (6), and `SVDNonconvergence` (7).
- LAPACK++ rejects negative `info` by throwing `lapack::Error`; positive `gesdd` info means nonconvergence. Operator, component, and allocation exceptions propagate unchanged.
- QB termination codes 0/2/3 produce a usable approximation and return `Success`; this does not promise the requested tolerance was met.

Output pointers and their existing data remain unchanged on failure. The documentation also specifies ownership and that rank/RNG state are not rolled back. QB allocation checks now preserve pointers needed for cleanup. Deterministic tests cover codes 4/5/6, exception propagation, nonnull output preservation, and usable QB2/QB3 results; positive `gesdd` info and allocator exhaustion were not fault-injected. The dense implementation is unchanged.

--- Astra

